PR #43, opened 2024-03-05
Fix BT.CPP upgrade

tonynajjar:


opennav_coverage/src/visualizer.cpp
```diff
@@ -46,6 +46,7 @@ void Visualizer::visualize(
     for (unsigned int i = 0; i != utm_path->poses.size(); i++) {
       utm_path->poses[i].pose.position.x += ref_pt.getX();
       utm_path->poses[i].pose.position.y += ref_pt.getY();
+      utm_path->poses[i].header.frame_id = GLOBAL_FRAME; // Necessary for Foxglove
```
SteveMacenski, line 49:
```suggestion
      utm_path->poses[i].header.frame_id = GLOBAL_FRAME;  // Necessary for Foxglove
```

SteveMacenski (approved):
@tonynajjar just add in the docs the transient local-ness of those topics

README.md
```diff
@@ -1,6 +1,6 @@
...
+This package contains the Complete Coverage Task Server & auxiliary tools utilizing the [Fields2Cover](https://github.com/Fields2Cover/Fields2Cover) complete coverage planning system which includes a great deal of options in headland, swath, route, and final path planning. You can find more information about Fields2Cover (F2C) in its [ReadTheDocs Documentation](https://fields2cover.github.io/index.html). It can accept both GPS and Cartesian coordinates and publishes the field, headland, swaths, and route as separate topics in cartesian coordinates for debugging and visualization (topics have Transient Local durability for late-joining visualization tools). It can also compute coverage paths based on open-field polygons **or** based on annotated rows as might exist in a tree farm or other applications with both irregular and regular pre-established rows.
```
tonynajjar, line 3: Do you have a better place for this comment?

SteveMacenski (approved):

SteveMacenski:
Linting failed an an odd error basically complaining about conflicts with every BT.CPPv4 node?

opennav_coverage/CMakeLists.txt
```diff
@@ -42,6 +43,7 @@ set(dependencies
   builtin_interfaces
   tf2_ros
   opennav_coverage_msgs
+  ament_index_cpp
```
tonynajjar, line 46: No idea why, but only in this package do it get:

```
/opt/overlay_ws/src/opennav_coverage/opennav_coverage/test/test_utils.cpp:21:10: fatal error: ament_index_cpp/get_package_share_directory.hpp: No such file or directory
   21 | #include "ament_index_cpp/get_package_share_directory.hpp"
```

although other packages have this include and don't explicitly have ament_index_cpp in CMakelist

tonynajjar:
> Linting failed an an odd error basically complaining about conflicts with every BT.CPPv4 node?

I found a couple of build issue that I fixed. The logs were super verbose though, maybe there is something else, let's see

tonynajjar:
It's still failing, looking at the logs, they contain both `/opt/ros/iron/include/behaviortree_cpp_v3` and `/opt/ros/iron/include/behaviortree_cpp`, maybe that's a problem

opennav_coverage/CMakeLists.txt
SteveMacenski: That is really odd. I didn't see that but maybe I need to break my own workspace cache to have seen that.

SteveMacenski:
That is odd - this CI shouldn't have any caching in it (@emersonknapp any caching happen in the ros action ci?) for it to know about BT.CPPv4 if its gone here. Does this cause you issues locally?

opennav_coverage/CMakeLists.txt
SteveMacenski: I'd say just to leave it - it doesn't hurt anyone.

tonynajjar:
> Does this cause you issues locally?

Locally on rolling no, I didn't test on iron though

tonynajjar:
> That is odd - this CI shouldn't have any caching in it (@emersonknapp any caching happen in the ros action ci?) for it to know about BT.CPPv4 if its gone here

@SteveMacenski I'm guessing it knows about BT.CPPv4 because it's using iron. I tried using rolling in this [commit](https://github.com/open-navigation/opennav_coverage/pull/43/commits/3669e7d84d3065b0a149a20babe35fe5eba008b8) but that fails as well because it can't find `nav2_util` -> are there no binaries for nav2 on rolling?

SteveMacenski:
There is no Nav2 binaries for rolling, but you could add it in the https://github.com/open-navigation/opennav_coverage/blob/main/.github/deps.repos for use in CI for testing. I'm not sure if you can isolate it to only nav2_utils and its necessary dependencies, but I suppose we could change the Action's `colcon build` to include `packages-up-to` for the opennav coverage packages, which would include only the parts of Nav2 strictly required. It would though need to build the Nav2 BT libraries which is non-trivial.

SteveMacenski:
-_- action ci tools don't call rosdep, I literally can't believe that

tonynajjar:
> -_- action ci tools don't call rosdep, I literally can't believe that

hmm I think it does
image

but for some reason it can't fint "standard" packages: https://github.com/open-navigation/opennav_coverage/actions/runs/8175554856/job/22352920043?pr=43

tonynajjar:
I'm wondering if https://discourse.ros.org/t/preparing-ros-2-rolling-for-the-transition-to-ubuntu-24-04/35673/8 could have anything to do with it. It still says Jammy in the CI though `No definition of [geometry_msgs] for OS version [jammy]`

tonynajjar:
> I'm wondering if https://discourse.ros.org/t/preparing-ros-2-rolling-for-the-transition-to-ubuntu-24-04/35673/8 could have anything to do with it. It still says Jammy in the CI though `No definition of [geometry_msgs] for OS version [jammy]`

Now I have a stronger suspicion:

```
root@tony-xmg-22:/opt/overlay_ws# rosdep resolve geometry_msgs
#apt
ros-rolling-geometry-msgs
root@tony-xmg-22:/opt/overlay_ws# ^C
root@tony-xmg-22:/opt/overlay_ws# ^C
root@tony-xmg-22:/opt/overlay_ws# ^C
root@tony-xmg-22:/opt/overlay_ws# ^C
root@tony-xmg-22:/opt/overlay_ws# ^C
root@tony-xmg-22:/opt/overlay_ws# rosdep update
reading in sources list data from /etc/ros/rosdep/sources.list.d
Warning: running 'rosdep update' as root is not recommended.
  You should run 'sudo rosdep fix-permissions' and invoke 'rosdep update' again without sudo.
Hit https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/osx-homebrew.yaml
Hit https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/base.yaml
Hit https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/python.yaml
Hit https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/ruby.yaml
Hit https://raw.githubusercontent.com/ros/rosdistro/master/releases/fuerte.yaml
Query rosdistro index https://raw.githubusercontent.com/ros/rosdistro/master/index-v4.yaml
Skip end-of-life distro "ardent"
Skip end-of-life distro "bouncy"
Skip end-of-life distro "crystal"
Skip end-of-life distro "dashing"
Skip end-of-life distro "eloquent"
Skip end-of-life distro "foxy"
Skip end-of-life distro "galactic"
Skip end-of-life distro "groovy"
Add distro "humble"
Skip end-of-life distro "hydro"
Skip end-of-life distro "indigo"
Add distro "iron"
Skip end-of-life distro "jade"
Skip end-of-life distro "kinetic"
Skip end-of-life distro "lunar"
Skip end-of-life distro "melodic"
Add distro "noetic"
Add distro "rolling"
updated cache in /root/.ros/rosdep/sources.cache
root@tony-xmg-22:/opt/overlay_ws# rosdep resolve geometry_msgs

ERROR: No definition of [geometry_msgs] for OS version [jammy]

No definition of [geometry_msgs] for OS version [jammy]
	rosdep key : geometry_msgs
	OS name    : ubuntu
	OS version : jammy
	Data:
_is_ros: true
		debian:
		  bookworm:
		    apt:
		      packages:
		      - ros-rolling-geometry-msgs
		osx:
		  homebrew:
		    packages:
		    - ros/rolling/common_interfaces
		rhel:
		  '9':
		    dnf:
		      packages:
		      - ros-rolling-geometry-msgs
		ubuntu:
		  noble:
		    apt:
		      packages:
		      - ros-rolling-geometry-msgs
```

SteveMacenski:
Oh I see that too. So I guess transient during the Rolling migration that's underway, I see the flood of build farm emails with their work to `24.04`. What's the path from here? I think this was just a test to see if it would resolve on rolling since Iron was problematic. Certainly we could move the CI to `rolling` too, I could _roll_ with that :wink:

SteveMacenski:
By the way, this is not just a CI problem, I see this locally as well after doing an apt update + upgrade using the BTv4 branch. I actually see this in opennav_docking as well now after the update which is interesting.

SteveMacenski:
Experimenting now on the odd BT compiling issues. Did you happen to make any progress on it? I'm starting with purging v3 from my computer so its only v4. I suspect that has something to do with it.

tonynajjar:
Mmm no progress as I didn't setup Iron. On my setup on rolling there wass only BTv4 so I couldn't reproduce locally. You were able to reproduce on iron right?

SteveMacenski:
I don't recall specifically what my setup was when I saw it, I was in the middle of doing something else so I just removed `opennav_coverage` and went back to what I was trying to do at the time when I ran into it. Its possible I had `nav2` main built on iron base (which could have had btv4 in it) or some other mixed policy. I was poking around alot of projects this week so I can't recall which specifically it was I found it with respect to. 

What I know I did right now:
- Rolling, with BTv4 installed only (apt removed v3)
- Built in a workspace with Nav2 clean with no prior caches

Doesn't happen for one of the projects, I adapted that here as well now too in a separate branch with just the v4 changes and that works too. I'll open a separate PR in a moment with just the v4 updates to see what CI says about that and that I have something working locally.

SteveMacenski:
Ah, in CI

> executing command [apt-get install -y ros-iron-nav2-behavior-tree]

That's going to pull in BT.CPPv3. Then we're using v4 that would explain it.

tonynajjar:
Yes in BTv3 is expected in Iron so the only solution I see is to port the CI to rolling when rosdep works again... Do you see another option?

SteveMacenski:
Note that you missed adding `BTCPP_format="4" ` to the BT XML + the inline test XML string which causes a warning on execution. Please add that and I'll close my PR (only difference) 

> I see is to port the CI to rolling when rosdep works again

I agree. If I'm reading the docs for `action-ros-ci` correctly, we can use `colcon-extra-args: --packages-up-to <packages in repo>` to automatically only build what is needed from Nav2. Else, we can specify exactly which packages to build with `package-name:`, but we'll need to figure out exactly which packages we need to build manually. Which, from a quick look, seems like ~46 packages which would be a pain in the ass to list. 

https://github.com/ros-tooling/action-ros-ci/blob/634e9ac5fb970763bd6207e31e1875d0081be71e/action.yml#L31-L34

tonynajjar:
> https://github.com/ros-tooling/action-ros-ci/blob/634e9ac5fb970763bd6207e31e1875d0081be71e/action.yml#L31-L34

ah nice find, there's nothing about it in the readme

> I think that's the path forward and the contingencies, do you agree?

yep sounds like a plan. Let's see what they'll answer in the discourse

 I'm looking forward to this moving in navigation2 (if that's gonna happen) to manage it more easily and avoid the double work.

SteveMacenski:
Yeah the plan is to move it to Nav2 once the F2C updated features are in F2C & updated to be used here.

tonynajjar:
I tried again since [rolling sync is out](https://discourse.ros.org/t/preparing-for-rolling-sync-2024-03-13/36625) but still same issue with rosdep not finding some packages. Unfortunately I don't have the time currently to look further into it

SteveMacenski:
It still failing everywhere, I wouldn't worry about it yet

PR #44, opened 2024-03-06
Test PR for CI [Do not merge]

tonynajjar:


tonynajjar:
This also fails but for a different reason than https://github.com/open-navigation/opennav_coverage/pull/43, it fails because of tests

SteveMacenski:
This is a known issue that I ran out of time in my contract to fully solve. It only appears in the GitHub Actions. The other issues is compiling, that's a bigger deal :wink:

Issue #64, opened 2024-07-03
Select start pose option


MarcM0:
I have done some visualization of the planned paths. It seems the path starts at an arbitrary point near the left of the polygon, (not the corner of the polygon where the bot is). Is there a way to control this? If not, is there a way to get the start pose before path execution starts so that we can navigate to it?

SteveMacenski:
An image of what you mean would be helpful and detail about which node you're using. The paths of the field now (not row node) should start at one of the corners of the field, minus the headlands and the tool's radius. So there are different variant options. 

v1.X in F2C includes the ability to modify the `variant` to use of those options in the API [1], however, there's not really a way to know which is which. So you'd basically just have to guess and check or generate all possible variants and then sort it out yourself as to which you want to use. For that reason, I chose not to expose that feature since it wasn't knowable beforehand & you could get the same effect just by ordering the points differently in your input file (i.e. roll the points' positions in the inputs). 

But, I did think to myself someone might eventually want that feature, so I didn't do anything that prevents a contributor from adding in which `variant` to use in the request field and propagating it to the route planner of choice. Or I suppose we could modify the request to add an option for `generate_all_variants` and return a vector of fields to choose from. 

[1] https://github.com/Fields2Cover/Fields2Cover/blob/v1.2.0/include/fields2cover/route_planning/single_cell_swaths_order_base.h#L22

> is there a way to get the start pose before path execution starts so that we can navigate to it?

Yeah! Look at the path and take its first point `path.poses[0].pose.position` and navigate there with Nav2 before then sending the rest of the plan.

MarcM0:
Thanks for the reply, I am essentially using the demo, [https://github.com/open-navigation/opennav_coverage/tree/iron/opennav_coverage_demo](url).

> Yeah! Look at the path and take its first point path.poses[0].pose.position and navigate there with Nav2 before then sending the rest of the plan. 
I mean through the python interface. All I see is 
```
goal_msg = NavigateCompleteCoverage.Goal()
goal_msg.frame_id = frame_id
goal_msg.polygons.append(self.toPolygon(field))
self.coverage_client.send_goal_async(goal_msg, self._feedbackCallback)
```
 Am I supposed to use a compute message first? If so, then how do I run the computed path?

> get the same effect just by ordering the points differently in your input file (i.e. roll the points' positions in the inputs).
I tried rotating the input array, but no matter what is starts at the point with the lowest x value.

SteveMacenski:
[...]

PR #65, opened 2024-07-04
Add behavior tree to navigate to start before starting fields navigation

MarcM0:
Added example behavior tree to navigate to the start of the field before navigating the field. See https://github.com/open-navigation/opennav_coverage/issues/64


MarcM0:
It doesn't look like this build issue is related to my commits

SteveMacenski (approved):

Issue #58, opened 2024-05-24
ROS2 Humble: Failed to build


grboguz21:
Hi, 
I'm using **_Ubuntu 22.04_** and _**ROS2 Humble**_. I'm following the steps below but can't build the humble branch:
```
git clone https://github.com/Fields2Cover/Fields2Cover.git
git clone https://github.com/open-navigation/opennav_coverage.git -b humble
colcon build --packages-select opennav_coverage opennav_coverage_demo opennav_coverage_msgs fields2cover

ros2 launch opennav_coverage_demo coverage_demo_launch.py
```

When I build the packages as above, I'm getting lots of error. The last part of the error is something like this:
```
gmake[2]: *** [CMakeFiles/opennav_coverage_core.dir/build.make:90: CMakeFiles/opennav_coverage_core.dir/src/headland_generator.cpp.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[2]: *** [CMakeFiles/opennav_coverage_core.dir/build.make:118: CMakeFiles/opennav_coverage_core.dir/src/route_generator.cpp.o] Error 1
gmake[2]: *** [CMakeFiles/opennav_coverage_core.dir/build.make:104: CMakeFiles/opennav_coverage_core.dir/src/swath_generator.cpp.o] Error 1
gmake[2]: *** [CMakeFiles/opennav_coverage_core.dir/build.make:146: CMakeFiles/opennav_coverage_core.dir/src/visualizer.cpp.o] Error 1
gmake[2]: *** [CMakeFiles/opennav_coverage_core.dir/build.make:132: CMakeFiles/opennav_coverage_core.dir/src/path_generator.cpp.o] Error 1
gmake[2]: *** [CMakeFiles/opennav_coverage_core.dir/build.make:76: CMakeFiles/opennav_coverage_core.dir/src/coverage_server.cpp.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:188: CMakeFiles/opennav_coverage_core.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2
---
Failed   <<< opennav_coverage [13.1s, exited with code 2]

Summary: 2 packages finished [1min 8s]
  1 package failed: opennav_coverage
  2 packages had stderr output: fields2cover opennav_coverage
  1 package not processed
```

Error Message (whole message):

[...]

PR #82, opened 2024-11-21
Update to Jazzy/Rolling with F2C v1.2.1

adivardi:
This PR updates the code to build and run on Jazzy/Rolling, but still with Fields2Cover v1.2.1.
It requires F2C with branch v1.2.1-devel` to work (https://github.com/Fields2Cover/Fields2Cover/tree/v1.2.1-devel)

The biggest change is the new BehaviorTree version, plus a few fixes for compiler errors.
I also updated the `coverage_demo_launch.py` and `row_coverage_demo_launch` to work with the new Gazebo and with `nav2_minimal_tb3_sim`.

It also updates the CI workflows to base on Ubuntu Noble and ROS Jazzy.

~~Also moved the initial robot pose and polygon to (0,0). In GZ Ignition, tb3 publishes its ground truth pose as odometry with respect to its start position. Even when spawning the model at (5,5), the map->base_link transform is 0 because the robot did not drive. This causes navigation to fail as the robot pose is too far from the start of the path.~~

Installation instructions: There is no need to manually build F2C (as in #58 ). Simply cloning it into the workspace with the correct branch and building everything with colcon works.


opennav_coverage/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.5)
 project(opennav_coverage)
 
 find_package(ament_cmake REQUIRED)
+find_package(ament_index_cpp REQUIRED)
```
SteveMacenski, line 5: I don't see where this is used

SteveMacenski:
> Also moved the initial robot pose and polygon to (0,0). In GZ Ignition, tb3 publishes its ground truth pose as odometry with respect to its start position. Even when spawning the model at (5,5), the map->base_link transform is 0 because the robot did not drive. This causes navigation to fail as the robot pose is too far from the start of the path.

But the pose is spawned at 5,5 - so it should be fine? I guess its not clear to me why this would need to have been changed given the demo's worked to-date

> TODO: update row_ launch as well

Will wait for that :-)

SteveMacenski:
Update https://github.com/open-navigation/opennav_coverage/blob/main/.github/workflows/test.yml#L31 to `v4` to get CI to run (also, check the linter)

opennav_coverage/CMakeLists.txt
adivardi: I added it because the tests were failing to building because of the line
```
#include "ament_index_cpp/get_package_share_directory.hpp"
```

It is in `test_server.cpp` and in `test_utils.cpp`

adivardi:
> I don't think you should need to manually go into the workspace to build F2C, it should work with colcon build in you workspace with the nav2 coverage servers -- that's how I developed this at least.

Oh you are right. I think I never tried it since it didn't build directly on rolling and then I was following the #58 instructions. I added a note on that.

SteveMacenski (changes requested):
There are some weird compilation errors in CI. Maybe we need to update to BT.CPP in Jazzy (so upgrade our CI in L19 of `test.yml` github workflow to be `jazzy` instead of `iron`? Check the CI job, but either way moving up to `jazzy` would be good since `iron` is EOL (and Jazzy wasn't around when we developed this CI).

Do the tests and such work for you locally? 

Minus this quirk, I approve of the rest :-)

SteveMacenski:
>  now it build fine but 2 tests are failing.

Its actually 1 test `testServerTransactions`, the output is just repeated twice for 2 different reasons. This looks like the issue

> ROR] [1732619322.500486728] [my_node.rclcpp_action]: unknown goal response, ignoring...

Or possibly the

>       [INFO] [1732619322.524461239] [coverage_server]: Generating coverage path in EPSG:4258 frame for zone with 13 outer nodes and 0 inner polygons.       -- run_test.py: return code -11

Whereas its segfaulting somewhere mid-execution with this update (possibly due to F2C issues). What if we use `v1.2.1` without the `-devel`? I've recently been in a company office that they tried that out and worked fine. The unknown goal response could be unrelated just due to the test structures (or not)

.github/deps.repos
```diff
@@ -2,4 +2,4 @@ repositories:
   Fields2Cover:
     type: git
     url: https://github.com/Fields2Cover/Fields2Cover.git
-    version: main
+    version: v1.2.1-devel
```
SteveMacenski, line 5:
```suggestion
    version: v1.2.1
```

.github/deps.repos
```diff
@@ -2,4 +2,4 @@ repositories:
   Fields2Cover:
     type: git
     url: https://github.com/Fields2Cover/Fields2Cover.git
-    version: main
+    version: v1.2.1
```
SteveMacenski, line 5:
```suggestion
    version: v1.2.1-devel
```

adivardi:
`v1.2.1` doesn't build on my laptop because of this errors. The `v1.2.1-devel` just fixes those, so it would be a bit strange if that was causing the issue, but can try that.

SteveMacenski:
Seems to mostly turn over - though perhaps the file replacement makes tinyxml unhappy for some reason? 

```
  [0.127s] WARNING:colcon.colcon_test_result.test_result.xunit:Skipping 'build/fields2cover/_deps/tinyxml2-src/resources/empty.xml': no element found: line 1, column 0
  [0.127s] WARNING:colcon.colcon_test_result.test_result.xunit:Skipping 'build/fields2cover/_deps/tinyxml2-src/resources/xmltest-4636783552757760.xml': not well-formed (invalid token): line 1, column 0
  [0.128s] WARNING:colcon.colcon_test_result.test_result.xunit:Skipping 'build/fields2cover/_deps/tinyxml2-src/resources/xmltest-5330.xml': syntax error: line 1, column 0
  [0.128s] WARNING:colcon.colcon_test_result.test_result.xunit:Skipping 'build/fields2cover/_deps/tinyxml2-src/resources/xmltest-5662204197076992.xml': not well-formed (invalid token): line 1, column 2
  [0.128s] WARNING:colcon.colcon_test_result.test_result.xunit:Skipping 'build/fields2cover/_deps/tinyxml2-src/resources/xmltest-5720541257269248.xml': not well-formed (invalid token): line 1, column 8
  [0.128s] WARNING:colcon.colcon_test_result.test_result.xunit:Skipping 'build/fields2cover/_deps/tinyxml2-src/resources/out/compact.xml': no element found: line 1, column 0
  [0.132s] WARNING:colcon.colcon_test_result.test_result.xunit:Skipping 'build/fields2cover/_deps/tinyxml2-src/resources/out/test7.xml': syntax error: line 3, column 0
```

opennav_coverage/CMakeLists.txt
```diff
@@ -19,7 +20,7 @@ find_package(Fields2Cover REQUIRED)
 
 # potentially replace with nav2_common, nav2_package()
 set(CMAKE_CXX_STANDARD 17)
-add_compile_options(-Wall -Wextra -Wpedantic -Werror -Wdeprecated -fPIC -Wshadow -Wnull-dereference)
+# add_compile_options(-Wall -Wextra -Wpedantic -Werror -Wdeprecated -fPIC -Wshadow -Wnull-dereference)
```
SteveMacenski, line 23: Add comment: "These are commented out due to quality issues compiling F2C"

adivardi: oh, I can revert that since switching to v1.2.1 didn't solve the issue anyway.

adivardi:
So the `error code -11` is only gone now that I switched away from the `EPSG:4258` coordinates file.
Maybe it still is still similar to https://github.com/Fields2Cover/Fields2Cover/issues/73#issuecomment-1736605501

SteveMacenski:
Its failing with: 

```
      /__w/opennav_coverage/opennav_coverage/ros_ws/src/iphd6oteara/opennav_coverage/opennav_coverage/test/test_server.cpp:119: Failure
       Expected equality of these values:
        result.code
          Which is: 1-byte object <06>
        rclcpp_action::ResultCode::SUCCEEDED
          Which is: 1-byte object <04>
```

I dont' think this is a timeout, but a problematic response code due to a failure. We see a log right before it that points to the issue at hand:

```
  Error: ROR] [1732792205.749924797] [coverage_server]: Invalid GML File or Coordinates: Coordinate system not recognized
```

adivardi:
OK, at this point the tests are passing, except `test_compute_coverage_path.cpp` which I disabled.

- Using `cartesian_test_field.xml` gave the error `Invalid GML File or Coordinates: Coordinate system not recognized` because the server was configured with `cartesian_frame_ = true;`. Setting it to `true` fixed the error (this is [this run](https://github.com/open-navigation/opennav_coverage/actions/runs/12116641334/job/33777421317?pr=82))
- switching back to `test_field.xml` passes now ([this run](https://github.com/open-navigation/opennav_coverage/actions/runs/12116949922/job/33778382418?pr=82)). I am not exactly sure why it was failing before...

adivardi (commented):
Finally, the tests are passing :) 

The last test was a timeout in `test_compute_coverage_path`, due to the server_thread hanging forever when trying to rejoin. un-commenting the `rclcpp::shutdown()` fixed that

SteveMacenski (approved):
Awesome, thanks so much for the hard work and patience to figure out the CI issues -- Merging now :partying_face: 

https://media.giphy.com/media/ta83CqOoRwfwQ/giphy.gif

adivardi:
@SteveMacenski Thanks!
Could you release this on Jazzy? right now there is no `ros-jazzy-opennav-coverage`

SteveMacenski:
Unfortunately due to the API instability in F2C and the binaries not necessarily working (perhaps fixed now?), I cannot release this to the build farm without it breaking regularly :( This is the main reason its still in `opennav_coverage` instead of integrated into Nav2 directly (i.e. `nav2_coverage`).

adivardi:
But it is released to iron and humble, is there a difference? or are those also breaking regularly?

SteveMacenski:
It is not released in any distribution https://build.ros2.org/search/?q=opennav_coverage&Jenkins-Crumb=c6c6671c175be7303334d165e6229c1395275a64b34e47dfb074fbb571abfe7d

Oh, do you mean a Jazzy branch? I can do that :-)

adivardi:
I guess I misremembered

For now we have to fix F2C to the `v1.2.1-devel` branch anyway, at least until #53 get some progress. So I think that's fine and probably a good practice anyway.
I think it would be great to not be forced to build opennav_coverage from source, though understand if this is not the highest as it builds well with colcon.

SteveMacenski:
For now we can't release binaries since things aren't necessarily stable, but we could instead vendorize (Example: https://github.com/SteveMacenski/spatio_temporal_voxel_layer) F2C at a static git commit hash so that new changes don't break v1.2.1 for our needs. That is not the perfered solution, but technically possible. We could have a vendorized F2C package in this repo that then the `opennav_coverage` relies on (instead of F2C directly) and I could actually release that

PR #99, opened 2026-01-23
Add frame_id to poses in visualizer

lodauria:
Adding frame_id to each point of the path inside the visualizer allows a correct visualization on FoxGlove Studio. Without it, it gives a warning and paths can't be visualized


SteveMacenski (approved):

SteveMacenski:
Thanks!

PR #103, opened 2026-03-26
Fix Broken isValid Check in Path Components Iterator

### LINKED ISSUE #102: Path Components Iterator Broken isValid() Check
The path components iterator provided in the [utils.hpp](./opennav_coverage/include/opennav_coverage/utils.hpp) has an incorrect validity check:
```c++
// line 371 of utils.hpp
/**
 * @brief For condition if still valid to continue iterating
 */
bool isValid()
{
  return idx_ <= max_idx_;
}
```

The max_idx_ is set to the size of the swaths vector, so to avoid accessing garbage values in the getNext() operation, the isValid check should just be checking idx_ < max_idx_.

Jquinny:
This is a really really simple PR to fix the isValid check #102 

closes #102

### DIFF
--- a/opennav_coverage/include/opennav_coverage/utils.hpp
+++ b/opennav_coverage/include/opennav_coverage/utils.hpp
@@ -372,7 +372,7 @@ public:
    */
   bool isValid()
   {
-    return idx_ <= max_idx_;
+    return idx_ < max_idx_;
   }
 
   /**
--- a/opennav_coverage/test/test_utils.cpp
+++ b/opennav_coverage/test/test_utils.cpp
@@ -231,6 +231,9 @@ TEST(UtilsTests, TestPathComponentsIterator)
     i++;
   }
 
+  // validate that our iteration value is equal to the size of the input swath
+  EXPECT_EQ(i, msg.swaths.size());
+
   auto last_row_info = it.getNext();
   EXPECT_EQ(std::get<1>(last_row_info), nullptr);
 }


SteveMacenski (approved):
The only thing I'd like is for you to confirm to me that you tested before this PR and confirmed the issue in its use in the code. I trust this is a valid solution but I want to make sure there's not some reason I did that which may not be obvious.

Why is the test not failing? https://github.com/open-navigation/opennav_coverage/blob/main/opennav_coverage/test/test_utils.cpp

Jquinny:
> The only thing I'd like is for you to confirm to me that you tested before this PR and confirmed the issue in its use in the code. I trust this is a valid solution but I want to make sure there's not some reason I did that which may not be obvious.
> 
> Why is the test not failing? https://github.com/open-navigation/opennav_coverage/blob/main/opennav_coverage/test/test_utils.cpp



> The only thing I'd like is for you to confirm to me that you tested before this PR and confirmed the issue in its use in the code. I trust this is a valid solution but I want to make sure there's not some reason I did that which may not be obvious.
> 
> Why is the test not failing? https://github.com/open-navigation/opennav_coverage/blob/main/opennav_coverage/test/test_utils.cpp

Sounds good. I've added a quick iteration variable test to the current path components iterator test function. When running the tests with the old isValid check, I get the following from the test log file:
```
9: [ RUN      ] UtilsTests.TestPathComponentsIterator
9: /home/user/ros2_ws/src/opennav_coverage/opennav_coverage/test/test_utils.cpp:235: Failure
9: Expected equality of these values:
9:   i
9:     Which is: 5
9:   msg.swaths.size()
9:     Which is: 4
9: 
9: [  FAILED  ] UtilsTests.TestPathComponentsIterator (0 ms)
```

This highlights that the iteration loop is going one past the expected amount. Once I added back in the quick fix the test passed

SteveMacenski (approved):

PR #105, opened 2026-06-24
Migrate to Fields2Cover v2 (Humble)

u33549:
## Summary

Migrates `opennav_coverage` and `opennav_row_coverage` from Fields2Cover 1.2.1 to **Fields2Cover v2** (verified against `v2.0.0`). All breaking API usage is updated so both packages build and pass their tests against F2C v2, with no intended behavioral change to the coverage pipeline.

## Motivation

Fields2Cover v2 introduced several breaking API changes: encapsulated data models (private members + accessors), renamed methods, and a restructured robot model. This brings the coverage server stack up to date with the current F2C release.

## Changes

Each commit is scoped to one migration area:

| Area | Change |
| --- | --- |
| `types.hpp` | Remove duplicate `LineString` typedef |
| Robot model | `F2CRobot` public fields → setters/getters (`setWidth`/`setCovWidth`/`setMinTurningRadius`/`setMaxDiffCurv`, `getWidth`/`getCovWidth`); same in both dynamic-parameter callbacks |
| Swath | `BruteForce::step_angle` field → `setStepAngle()` |
| Path | `PathPlanning::searchBestPath` → `planPath`; `turn_point_dist` field → `TurningBase::setDiscretization()` |
| Coverage servers | `F2CField` `coord_sys`/`field` members → `getCRS()`/`getField()` |
| `utils.hpp` | `Path::states` → `operator[]`/`size()`/`back()` (now private) |
| Row swath | `LineString::StartPoint()/EndPoint()` → camelCase `startPoint()/endPoint()` |
| Tests | Updated to v2 API; fixed `generateRandField` argument order (swapped in v2) and `PathState::type` now defaulting to `SWATH` |

Headland and route generators required no source changes (their APIs are unchanged in v2).

### Build system

- Link `GDAL::GDAL` explicitly — v2's CMake target no longer propagates GDAL as an interface dependency, so consumers emitting OGR symbols inline must link it themselves.
- Resolve tinyxml2 in `opennav_row_coverage` via `tinyxml2_vendor` + `find_package(TinyXML2)` (the system `libtinyxml2-dev` ships no CMake config).
- Select the F2C target by name (`Fields2Cover` for source installs, `Fields2Cover::Fields2Cover` for the ROS/apt package) and mark F2C headers as SYSTEM includes so their own `-Wshadow` warnings do not break this `-Werror` build.

## Validation

- Both packages build against F2C `v2.0.0` (tested with the ROS apt package `ros-humble-fields2cover` and a from-source install).
- All unit tests pass: **207 tests, 0 failures**.
- End-to-end check in RViz: the full headland → swath → route → path pipeline produces a correct boustrophedon coverage plan with Dubins turns.

## Behavioral validation (RViz)

_Coverage plan for a 50×50 m cartesian field: outer boundary (green), headland-inset planning field (orange), parallel swaths, and connecting Dubins turns._

## Notes

- Verified against the F2C `v2.0.0` tag. Behavioral parity with v1 (e.g. turn-point density via `setDiscretization` vs the old `turn_point_dist`) was sanity-checked visually; a deterministic golden-output test would be a good follow-up.


u33549:
Note: this PR also includes small CI fixes (upload-artifact/checkout action bumps + pinning F2C to v2.0.0 in deps.repos) — humble's build job was failing on the now-disabled upload-artifact@v1 before any build ran. Happy to split these into a separate PR if you'd prefer.

u33549:
Also worth noting: Fields2Cover v2 is now available as a binary package — apt-get install ros-humble-fields2cover (currently 2.0.0) — so a from-source install is no longer required for local development. rosdep resolves the fields2cover key to it directly, and both packages build and pass tests against the apt package as well as a from-source build. CI still pins F2C to v2.0.0 from source in deps.repos for reproducibility.

SteveMacenski (approved):

SteveMacenski:
Can we update the `main` branch with this? The problem with updating the Humble branch is that we're essentially breaking API/ABI mid-distribution which is not good

u33549:
> Can we update the `main` branch with this? The problem with updating the Humble branch is that we're essentially breaking API/ABI mid-distribution which is not good

Thanks for the direction, a few thoughts:

1. I agree on avoiding an API/ABI break on humble — I'll retarget this to main. I'll need to validate the build and tests against Jazzy/Rolling first, but I'll move as fast as I can.

2. In the meantime, would it make sense to open a short-lived humble-v2 branch for users who want to opt in to F2C v2 early? The README would need updating either way — currently it still references v1.2.1 as the supported version and lists a from-source install, but ros-humble-fields2cover (v2.0.0) is now available as a binary apt package, so that section is already outdated.

Happy to work on the main migration together — just point me in the right direction if there's anything specific you'd like done differently for Jazzy/Rolling, and I'll get it done.

SteveMacenski:
Sure thing - see https://github.com/open-navigation/opennav_coverage/tree/humble-v2

u33549:
Cool, I've retargeted the base to humble-v2 and updated the README accordingly. Good to merge whenever you'd like.

SteveMacenski:
I'll merge this when paired with the `main` PR. I don't want that to get lost since then this would only be possible in Humble (which is starting to approach EOL) and not in newer distros like Jazzy and Lyrical. At the very least I want a `main` PR before merging humble as a bookkeeping mechanism

u33549:
Sounds good. I just started working on the Jazzy migration — I'll keep you posted. The plan is the same as humble: first resolve any API breaking changes for main, validate on both Jazzy and Rolling, then we can plan for new features from there.

u33549:
@SteveMacenski quick question —does main currently target Rolling only, or both Jazzy and Rolling? I'm running into some issues trying to support both and want to make sure I'm targeting the right distributions.

Update: I've got it building on both Jazzy and Rolling now.

u33549:
My bad on the update above — that was entirely on me. I'd assumed `main` covered both distros as a single shared branch. The reason my local runs looked green is that I wasn't rebuilding `behaviortree_cpp` / the BT pieces (I hadn't touched those parts), so the version mismatch never surfaced locally — my tests passed in my Jazzy environment and on `main` for both Jazzy and Rolling. It only showed up once CI built everything from source.

The actual failure on Jazzy is in `nav2_behavior_tree`:

```
loop_rate.hpp:64:27: error: 'class BT::Tree' has no member named 'wakeUpSignal'; did you mean 'emitWakeUpSignal'?
```

That looks like the newer `LoopRate` on `main` expecting a `behaviortree_cpp` version that's ahead of what Jazzy ships as a binary. Since there's a dedicated `jazzy` branch, my read now is: `main` targets Rolling, and Jazzy should track the `jazzy` branch instead. I'll pin `navigation2` per distro instead — `main` for Rolling, `jazzy` for Jazzy. That should fix the build on both. Once CI is green, I'll open the PRs against `main` and `jazzy` shortly.

u33549:
Hi @SteveMacenski — sorry for the frequent messages.

**Good news first:** inside dev containers, `opennav_coverage` + `opennav_row_coverage` build and pass **all tests** against Fields2Cover v2 on Jazzy, on the `ros-rolling-*` packages (Rolling on Noble), and on the `ros-lyrical-*` packages (Lyrical on Ubuntu 26.04). So the migration itself is in good shape.

**The wall I'm now hitting on Rolling CI:**
Rolling has moved off Ubuntu Noble (24.04) to its released form **Lyrical on Ubuntu 26.04 (resolute)** — the `ros-lyrical-*` packages are the current set, while the `ros-rolling-*` Noble packages are now frozen. Our CI was pinned to the `ubuntu-noble-ros-rolling` image, where:

- `rosdep` refuses to resolve rolling keys (`No definition of [ament_cmake] for OS version [noble]`), since rolling's rosdistro now targets 26.04, and
- the leftover `ros-rolling-*` Noble packages are frozen/ABI-skewed (missing `rosidl_buffer` targets, etc.).

**What I tried (chronologically):**

1. Hand-installing the missing nav2 source-build deps via apt on Noble → whack-a-mole + ABI skew.
2. Switched CI to the **Lyrical/resolute** `setup-ros-docker` image → rosdep works again. 
3. On fresh Lyrical, building the full nav2 stack from source surfaces ecosystem gaps not yet ported to 26.04:
   - `slam_toolbox` isn't in the rolling/lyrical rosdistro at all (only jazzy/kilted) — pulled in via `nav2_bringup`.
   - `ortools_vendor` has no Lyrical binary. I added both to `deps.repos` as source builds.
4. Ubuntu 26.04 ships **CMake 4**, which hard-errors on old `cmake_minimum_required(<3.5)` → fixed with `CMAKE_POLICY_VERSION_MINIMUM=3.5`.
5. After that, `ortools_vendor` (OR-Tools 9.9.0) **fails to compile on GCC 15**:
   ```
   scip-src/src/tinycthread/tinycthread.h:453: error: conflicting types for 'pthread_once_t'; have '__once_flag'
   ```
   This looks like an upstream OR-Tools/SCIP incompatibility with the 26.04 toolchain, and it would affect Fields2Cover too.

I've worked through several layers of this, but I'm honestly unsure what shape you'd want the CI to take here — so I'd really appreciate your guidance on the right approach before I keep pushing in a particular direction.

I'm still actively working on it in the meantime, and happy to take whatever direction you prefer.

u33549:
Opened the `main` PR to pair with this one: #110. It targets Rolling/Lyrical, brings the full stack (incl. `opennav_coverage_bt`/`opennav_coverage_navigator`) to Fields2Cover v2, and is green on Lyrical CI (272 tests, 0 failures). It also resolves the Lyrical/OR-Tools CI questions I raised above — details and a few CI sustainability caveats are in the PR description, would appreciate your guidance there.

SteveMacenski:
> does main currently target Rolling only, or both Jazzy and Rolling? 

Main targets rolling only. Jazzy or others are 'nice to have' but not required. Did you figure out your issues?

PR #109, opened 2026-06-28
Migrate to Fields2Cover v2 (Jazzy)

u33549:
## Summary

Migrates `opennav_coverage` and `opennav_row_coverage` from Fields2Cover 1.2.1 to Fields2Cover v2 (verified against v2.0.0) on the `jazzy` branch. All breaking API usage is updated so both packages build and pass their tests against F2C v2, with no intended behavioral change to the coverage pipeline.

This is the `jazzy` counterpart to the `humble-v2` migration (PR #105), ported via cherry-pick (zero conflicts — the `jazzy` branch shares the same `nav2_util`-based API as `humble`).

## Motivation

Fields2Cover v2 introduced several breaking API changes: encapsulated data models (private members + accessors), renamed methods, and a restructured robot model. This brings the coverage server stack up to date with the current F2C release on the Jazzy distribution.

## Changes

Each commit is scoped to one migration area:

| Area | Change |
|---|---|
| `types.hpp` | Remove duplicate LineString typedef |
| Robot model | `F2CRobot` public fields → setters/getters (`setWidth`/`setCovWidth`/`setMinTurningRadius`/`setMaxDiffCurv`, `getWidth`/`getCovWidth`); same in both dynamic-parameter callbacks |
| Swath | `BruteForce::step_angle` field → `setStepAngle()` |
| Path | `PathPlanning::searchBestPath` → `planPath`; `turn_point_dist` field → `TurningBase::setDiscretization()` |
| Coverage servers | `F2CField` `coord_sys`/`field` members → `getCRS()`/`getField()` |
| `utils.hpp` | `Path::states` → `operator[]`/`size()`/`back()` (now private) |
| Row swath | `LineString::StartPoint()`/`EndPoint()` → camelCase `startPoint()`/`endPoint()` |
| Tests | Updated to v2 API; fixed `generateRandField` argument order (swapped in v2) and `PathState::type` now defaulting to `SWATH` |

Headland and route generators required no source changes (their APIs are unchanged in v2).

## Build system

- Link `GDAL::GDAL` explicitly — v2's CMake target no longer propagates GDAL as an interface dependency.
- Resolve `tinyxml2` in `opennav_row_coverage` via `tinyxml2_vendor` + `find_package(TinyXML2)` (the system `libtinyxml2-dev` ships no CMake config).
[...]

PR #110, opened 2026-06-28
Fields2Cover v2 migration and fixes (Main - Rolling)

u33549:
## Summary

Brings the full `opennav_coverage` stack to **Fields2Cover v2** (verified against `v2.0.0`) on `main`, **and** fixes a number of build, runtime, and CI issues uncovered along the way. All breaking F2C API usage is updated and the build/CI is reworked so the whole package set — including `opennav_coverage_bt` and `opennav_coverage_navigator` — builds and passes tests on ROS 2 Rolling (released as **Lyrical** on Ubuntu 26.04). The coverage pipeline itself has no intended behavioral change.

## Motivation

Fields2Cover v2 introduced several breaking API changes: encapsulated data models (private members + accessors), renamed methods, and a restructured robot model. This brings the coverage server stack up to date with the current F2C release on the rolling/Lyrical distribution — and fixes the issues that surfaced when building and testing the full stack there.

## Changes — Fields2Cover v2 API

Each commit is scoped to one migration area (shared with the humble/jazzy PRs):

| Area | Change |
|---|---|
| `types.hpp` | Remove duplicate `LineString` typedef |
| Robot model | `F2CRobot` public fields → setters/getters (`setWidth`/`setCovWidth`/`setMinTurningRadius`/`setMaxDiffCurv`, `getWidth`/`getCovWidth`); same in both dynamic-parameter callbacks |
| Swath | `BruteForce::step_angle` field → `setStepAngle()` |
| Path | `PathPlanning::searchBestPath` → `planPath`; `turn_point_dist` field → `TurningBase::setDiscretization()` |
| Coverage servers | `F2CField` `coord_sys`/`field` members → `getCRS()`/`getField()` |
| `utils.hpp` | `Path::states` → `operator[]`/`size()`/`back()` (now private) |
| Row swath | `LineString::StartPoint()`/`EndPoint()` → camelCase `startPoint()`/`endPoint()` |
| Tests | Updated to v2 API; fixed `generateRandField` argument order (swapped in v2) and `PathState::type` now defaulting to `SWATH` |

Headland and route generators required no source changes (their APIs are unchanged in v2).

## Fixes

- **Out-of-bounds swath access (runtime correctness):** guard `PathComponentsIterator::getNext` so iterating path components no longer reads past the end of the swath collection.
- **`opennav_coverage_demo` reported as failing with no tests run:** its test dependency was declared with the legacy `tests_require`, which modern setuptools drops; colcon then fell back to the unittest runner, collected 0 tests, and exited with code 5. Declaring it via `extras_require={'test': ['pytest']}` makes colcon discover and run the package's `ament_lint` tests (copyright, flake8, pep257).
- **Deprecation fixes for Rolling:** `rclcpp::spin_some` → `SingleThreadedExecutor` in the viz test; both `get_package_share_directory` overloads are deprecated on Rolling but `get_package_share_path` does not exist on Jazzy, so the single-arg call is kept and wrapped in a `-Wdeprecated-declarations` pragma for one cross-distro code path.

## Changes — build system

- **`ament_target_dependencies` → `target_link_libraries`**: this macro was removed on Rolling, so `opennav_coverage`, `opennav_coverage_bt`, `opennav_coverage_navigator` and their tests now link modern namespaced targets (`pkg::pkg`, `${pkg_TARGETS}` for message packages, `nav2_util::nav2_util_core`, `nav2_ros_common::nav2_ros_common`, etc.).
- **GDAL**: link `GDAL::GDAL` explicitly — v2's CMake target no longer propagates GDAL as an interface dependency.
- **tinyxml2**: resolve in `opennav_row_coverage` via `tinyxml2_vendor` + `find_package(TinyXML2)` (the system `libtinyxml2-dev` ships no CMake config).
- **F2C target/headers**: select the F2C target by name (`Fields2Cover` for source installs, `Fields2Cover::Fields2Cover` for the packaged target) and mark F2C headers as `SYSTEM` so their own `-Wshadow` warnings do not break this `-Werror` build.

## Changes — CI (detailed)

The `Test` workflow is rebuilt to run on Lyrical with a manual colcon pipeline:

- **Runner image** is `setup-ros-docker ... resolute-ros-lyrical` (Ubuntu 26.04). ROS Rolling has been released as Lyrical there, where the apt repo is consistent and `rosdep` works — unlike rolling-on-Noble, where rosdep cannot resolve rolling keys.
- **Source dependencies** (`.github/deps.repos`), since nav2 and F2C are not published as binaries on Lyrical:
  - `navigation2` → `main`
  - `Fields2Cover` → `v2.0.0`
  - `ortools_vendor` → `main` (sets `USE_SCIP=OFF`, avoiding the SCIP/tinycthread GCC-15 failure)
  - `slam_toolbox` → `ros2`
  - `behaviortree_cpp` → `master` (nav2 `main` uses its unreleased `wakeUpSignal()` getter)
- **Manual steps instead of `action-ros-ci`**, because nav2 `main` needs a source patch applied between `vcs import` and `colcon build`:
  - Strip `-Werror` from `nav2_common`'s `nav2_package.cmake` (nav2 `main` is not `-Werror`-clean under GCC 15: deprecated `std::atomic_load` on `shared_ptr`, etc.). Our own opennav packages keep their `-Werror`.
  - Install system deps `rosdep` can't provide here: `libompl-dev` (for `nav2_smac_planner`'s `find_package(ompl)`), GDAL/Eigen/GEOS/tinyxml2, `tinyxml2-vendor`, Cyclone DDS.
  - **Drop shadowing apt packages** after rosdep:
    - `ros-lyrical-behaviortree-cpp` — the binary 4.9.0 would shadow the source-built `master` in the shared include tree.
    - `ros-lyrical-ompl` — rosdep resolves the `ompl` key to **2.0.1**, whose ReedsShepp API differs (`ReedsSheppStateSpace::ReedsSheppPath` / `reedsShepp()` renamed), breaking `nav2_smac_planner`. Removing it falls back to system `libompl-dev` **1.6.0**, the API nav2 `main` expects.
- **`CMAKE_POLICY_VERSION_MINIMUM=3.5`** — Ubuntu 26.04 ships CMake 4, which hard-errors on dependencies declaring `cmake_minimum_required(VERSION < 3.5)` (e.g. the zlib bundled by OR-Tools); this also propagates into nested FetchContent/ExternalProject sub-builds.
- **GCC 15 pinned** (`CC`/`CXX`) so the CI toolchain matches the dev container and rosdep cannot transitively pull GCC 16.
- **colcon workspace cache** — the third-party sources are imported into `src/deps` and cached together with `build/` and `install/`, keyed on `deps.repos`. They build from source once (~1 h) and are reused on every subsequent run (CI then finishes in minutes), which also removes a flaky OR-Tools parallel-build race (`gurobi.pb.h` generated late under `-j`). The cache is saved only after a successful build, so a broken/partial build is never cached; on a hit, the import and `-Werror` patch are skipped so restored sources keep their mtimes.
- **Scope**: the build covers `nav2_bt_navigator nav2_behavior_tree` plus all `opennav_*` packages; the test step runs the `opennav_*` suites.

`.github/lint.yml` also bumps `actions/checkout` to v4 and adds `opennav_row_coverage` to the lint matrix.

## Validation

- Full broad-scope build + tests are **green on Lyrical** (Test + Lint workflows): **272 tests, 0 failures, 44 skipped**.
- Reproduced end-to-end in a local Lyrical dev container that mirrors the CI exactly (persistent colcon cache).
- Behavioral parity with v1 was sanity-checked visually in RViz (the full headland → swath → route → path pipeline produces a correct boustrophedon plan with Dubins turns); a deterministic golden-output test would be a good follow-up.

## Notes

- The README's "Fields2Cover Installation" section is updated: `main` now uses F2C v2 from source (F2C has no binary on Lyrical, so rosdep cannot resolve the `fields2cover` key there).
- **Distribution targeting:** this PR targets **Rolling / Lyrical**. Jazzy is handled by the separate `jazzy` branch (PR #109) — nav2 `main`'s `LoopRate` uses `behaviortree_cpp`'s `wakeUpSignal()`, which is ahead of the `behaviortree_cpp` binary Jazzy ships (`'class BT::Tree' has no member named 'wakeUpSignal'`), as discussed in #105.
- **Resolves the Lyrical CI questions raised in #105:** the OR-Tools build failure on the 26.04 toolchain (`tinycthread.h` / conflicting `pthread_once_t`) is gone because `ortools_vendor`'s `main` sets `USE_SCIP=OFF`; `slam_toolbox` and `ortools_vendor` are built from source as they have no Lyrical binary; and CMake 4's hard error on old minimums is handled via `CMAKE_POLICY_VERSION_MINIMUM=3.5`. Building the full nav2 stack from source on Lyrical is what surfaced (and now verifies) the broad-scope build.

## CI caveats — interim approach

This CI gets a fully green broad-scope build on Lyrical today, but it is a pragmatic interim approach and may not be sustainable long-term. I want to be upfront about why (continuing from #105):

- It builds the **entire nav2 stack from source** (nav2 `main`, `slam_toolbox`, `ortools_vendor`, `behaviortree_cpp`). This is heavy and slow, and tightly couples our CI to the current state of those upstreams.
- It pins **moving branches** (`navigation2 main`, `behaviortree_cpp master`, `ortools_vendor main`) because that is what nav2 `main` requires right now. These are not reproducible — an upstream change can break our CI with no change on our side.
- It **drops shadowing apt packages** (`ros-lyrical-ompl` 2.0.1, `ros-lyrical-behaviortree-cpp` 4.9.0) and falls back to system/source versions. This is a workaround for upstream version skew on Lyrical, not a real fix, and will likely need revisiting as the Lyrical ecosystem stabilizes.

I went this route to unblock the migration and get a verifiable green build; would like to hear your guidance on how you'd like CI maintained here.


opennav_row_coverage/test/CMakeLists.txt
```diff
@@ -1,10 +1,11 @@
+# Each test links ${library_name}, which PUBLIC-links all dependency targets
+# (see ../CMakeLists.txt), so the dependencies propagate transitively. Rolling
+# removed ament_target_dependencies, hence the direct linking.
+
```
SteveMacenski: Remove

opennav_coverage/test/CMakeLists.txt
```diff
@@ -1,10 +1,11 @@
+# Each test links ${library_name}, which PUBLIC-links all dependency targets
+# (see ../CMakeLists.txt), so the dependencies propagate transitively. Rolling
+# removed ament_target_dependencies, hence the direct linking.
+
```
SteveMacenski: Remove AI generated messaging

opennav_coverage/test/test_server.cpp
```diff
@@ -97,8 +99,16 @@ TEST(ServerTest, testServerTransactions)
 
   auto goal_msg = opennav_coverage_msgs::action::ComputeCoveragePath::Goal();
   goal_msg.use_gml_file = true;  // Use file
-  goal_msg.gml_field =
-    ament_index_cpp::get_package_share_directory("opennav_coverage") + "/test_field.xml";
+  // get_package_share_directory(string) is the only overload present on all of
+  // humble, jazzy and rolling. Its replacement get_package_share_path() does
+  // not exist on humble/jazzy, so we keep using it and locally silence the
+  // deprecation warning that rolling emits (otherwise -Werror fails the build).
```
SteveMacenski: Remove AI generated messaging

opennav_coverage/CMakeLists.txt
```diff
@@ -46,6 +49,28 @@ set(dependencies
   opennav_coverage_msgs
 )
 
+# Modern CMake targets for the dependencies above. Rolling removed the
+# ament_target_dependencies macro, so we link these targets directly; this also
+# works on jazzy, which still exports the same targets. Message/interface
+# packages expose a ${pkg_TARGETS} variable, regular packages a pkg::pkg target.
```
SteveMacenski: Remove AI generated messaging

opennav_coverage_bt/CMakeLists.txt
```diff
@@ -38,13 +38,32 @@ set(dependencies
   behaviortree_cpp
 )
 
+# Modern CMake targets for the dependencies above. Rolling/Lyrical removed the
+# ament_target_dependencies macro, so we link these targets directly; this also
+# works on jazzy. Message packages expose a ${pkg_TARGETS} variable, regular
+# packages a pkg::pkg target. (nav2_bt_navigator is not used by the sources, so
+# it is exported as a dependency but not linked here.)
```
SteveMacenski: Remove AI generated messaging

opennav_coverage_navigator/CMakeLists.txt
```diff
@@ -34,12 +37,31 @@ set(dependencies
   opennav_coverage_msgs
 )
 
+# Modern CMake targets for the dependencies above. Rolling/Lyrical removed the
+# ament_target_dependencies macro, so we link these targets directly; this also
+# works on jazzy. Message packages expose a ${pkg_TARGETS} variable, regular
+# packages a pkg::pkg target.
+set(dep_targets
```
SteveMacenski: Remove AI generated messaging

opennav_row_coverage/test/test_swath_generator.cpp
```diff
@@ -38,9 +38,15 @@ class SwathTestFixture : public ::testing::Test
     node_ = std::make_unique<rclcpp::Node>("test_node");
     generator_ = std::make_unique<RowSwathGenerator>(node_);
 
+    // get_package_share_directory(string) is deprecated on rolling; locally
+    // silence the warning so -Werror does not fail the build (kept for jazzy
+    // compatibility, where its get_package_share_path() replacement is absent).
+#pragma GCC diagnostic push
```
SteveMacenski: Remove AI generated messaging

opennav_row_coverage/test/test_utils.cpp
```diff
@@ -33,18 +33,27 @@ namespace opennav_row_coverage
 
 TEST(UtilsTests, TestRowParserSimple)
 {
+  // get_package_share_directory(string) is deprecated on rolling; locally
+  // silence the warning so -Werror does not fail the build (kept for jazzy
+  // compatibility, where its get_package_share_path() replacement is absent).
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
SteveMacenski, line 38: Remove AI generated messaging

opennav_row_coverage/CMakeLists.txt
```diff
@@ -37,29 +41,65 @@ set(dependencies
...
 
+# Modern CMake targets for the dependencies above. Rolling removed the
+# ament_target_dependencies macro, so we link these targets directly; this also
+# works on jazzy, which still exports the same targets. Message/interface
+# packages expose a ${pkg_TARGETS} variable, regular packages a pkg::pkg target.
+# opennav_coverage uses the classic ament_export_libraries (no namespaced
+# target), so it is consumed via ${opennav_coverage_LIBRARIES}/_INCLUDE_DIRS
+# below. tf2_ros and the message packages are pulled in transitively by
+# find_package(opennav_coverage) through its ament_export_dependencies.
+set(dep_targets
```
SteveMacenski: Remove AI generated messaging

opennav_row_coverage/CMakeLists.txt
```diff
@@ -37,29 +41,65 @@ set(dependencies
...
+  tinyxml2::tinyxml2
+  GDAL::GDAL
 )
+target_include_directories(${library_name} PUBLIC ${opennav_coverage_INCLUDE_DIRS})
 
-target_link_libraries(${library_name} Fields2Cover tinyxml2::tinyxml2)
+# Treat Fields2Cover headers as SYSTEM so their own warnings do not break this
+# project's -Werror build. Uses the config-provided include variable (path- and
+# target-name-agnostic; apt /opt/ros or source /usr/local).
+if(Fields2Cover_INCLUDE_DIR)
```
SteveMacenski: Remove AI generated messaging

README.md
```diff
@@ -1,5 +1,7 @@
 # Open Navigation's Nav2 Complete Coverage
 
+> **Note:** This branch migrates `main` to **Fields2Cover v2** (verified against `v2.0.0`), replacing the previous Fields2Cover 1.2.1 dependency. See the companion `humble-v2` branch and the `jazzy` branch for the same upgrade on ROS 2 Humble and Jazzy.
```
SteveMacenski:
```suggestion
> **Note:** This branch migrates `main` to **Fields2Cover v2** (verified against `v2.0.0`), replacing the previous Fields2Cover 1.2.1 dependency. See the companion `humble-v2` branch and the `jazzy-v2` branch for the same upgrade on ROS 2 Humble and Jazzy.
```

.github/deps.repos
```diff
@@ -2,9 +2,33 @@ repositories:
...
     type: git
     url: https://github.com/ros-navigation/navigation2.git
     version: main
+
+  # Not yet released for Lyrical/Ubuntu 26.04, so build from source.
+  # ortools_vendor is a Fields2Cover dependency; slam_toolbox is pulled in by
+  # nav2_bringup. Both are absent from the lyrical apt repo (fresh distro).
+  # 'main' branch (not tag 9.9.0): USE_SCIP=OFF removes the SCIP/tinycthread
+  # build that fails on Ubuntu 26.04's GCC 15, and forwards
+  # CMAKE_POLICY_VERSION_MINIMUM=3.5 into the OR-Tools sub-build. PR #8.
```
SteveMacenski: Compress this into 1 line

.github/workflows/test.yml
```diff
@@ -11,25 +11,147 @@ jobs:
...
+      # it propagates into nested FetchContent/ExternalProject sub-builds too.
+      CMAKE_POLICY_VERSION_MINIMUM: "3.5"
+      # Pin GCC 15 (Ubuntu 26.04's default, and what the dev container uses, 15.2.0)
+      # so CI compiles with the same toolchain we validate against locally and
+      # rosdep cannot silently swap in gcc-16 as a transitive dependency.
+      CC: gcc-15
+      CXX: g++-15
     container:
-      image: ghcr.io/ros-tooling/setup-ros-docker/setup-ros-docker-ubuntu-noble-ros-rolling-ros-base:master
+      image: ghcr.io/ros-tooling/setup-ros-docker/setup-ros-docker-ubuntu-resolute-ros-lyrical-ros-base:master
```
SteveMacenski: This should use rolling

.github/workflows/test.yml
```diff
@@ -11,25 +11,147 @@ jobs:
...
+      # Ubuntu 26.04 ships CMake 4.x, which hard-errors on dependencies that
+      # declare cmake_minimum_required(VERSION < 3.5) (e.g. the zlib bundled by
+      # ortools_vendor's OR-Tools). This lets CMake 4 accept those old minimums;
+      # it propagates into nested FetchContent/ExternalProject sub-builds too.
+      CMAKE_POLICY_VERSION_MINIMUM: "3.5"
+      # Pin GCC 15 (Ubuntu 26.04's default, and what the dev container uses, 15.2.0)
+      # so CI compiles with the same toolchain we validate against locally and
+      # rosdep cannot silently swap in gcc-16 as a transitive dependency.
+      CC: gcc-15
+      CXX: g++-15
```
SteveMacenski: What's going on here

.github/workflows/test.yml
```diff
@@ -11,25 +11,147 @@ jobs:
...
     container:
-      image: ghcr.io/ros-tooling/setup-ros-docker/setup-ros-docker-ubuntu-noble-ros-rolling-ros-base:master
+      image: ghcr.io/ros-tooling/setup-ros-docker/setup-ros-docker-ubuntu-resolute-ros-lyrical-ros-base:master
     strategy:
       fail-fast: false
       matrix:
-          ros_distro: [rolling]
+          ros_distro: [lyrical]
```
SteveMacenski: Rolling

SteveMacenski (commented):
I don't understand what's happeningi n the test.yml file from a quick glance, but I think alot about it should be reverted. It looks like AI was pointed at it and added a ton of things that should be removed. Lets keep this simple. 

I'm also creating a `lyrical` branch if you want to have this target lyrical since it seems like you did that already! We can have Lyrical support V2 out of the box

.github/workflows/test.yml
u33549: These are two build-environment workarounds for the Lyrical (Ubuntu 26.04) toolchain.

**`CMAKE_POLICY_VERSION_MINIMUM=3.5`** — Ubuntu 26.04 ships CMake 4, which now *hard-errors* (no longer just warns) on any dependency that declares `cmake_minimum_required(VERSION < 3.5)`. That breaks the from-source `ortools_vendor` build — the zlib OR-Tools pulls in via FetchContent still declares an old minimum — so configure fails before anything compiles. This makes CMake 4 accept those old minimums, and it propagates into the nested FetchContent/ExternalProject sub-builds. It only matters because we build OR-Tools/F2C from source here; with binary deps it wouldn't be needed.

**`CC`/`CXX=gcc-15`** — just a reproducibility pin (15 is 26.04's default and what our dev container uses), and it stops rosdep from transitively pulling gcc-16 into the build.

More broadly, both exist only because this PR builds the nav2 stack + F2C/OR-Tools from source on Lyrical (no binaries there yet) — which is the CI-shape question I flagged in the description.

SteveMacenski: Got it, remove the gcc-15 pin, we should let it float by what's used in colcon for a given distribution. We'll forget to update it later in a few years time when that changes. 

Got it on the 3.5, that is annoying. We should update all our stuff to 3.8 but if ortools fails then I suppose that's a fine temp workaround

.github/deps.repos
```diff
@@ -2,9 +2,28 @@ repositories:
...
+  ortools_vendor:
+    type: git
+    url: https://github.com/Fields2Cover/ortools_vendor.git
+    version: main
+
+  slam_toolbox:
+    type: git
+    url: https://github.com/SteveMacenski/slam_toolbox.git
+    version: ros2
+
```
SteveMacenski: Just do `--packages-skip slam_toolbox`. There's no reason to need this here

SteveMacenski (commented):
CI still seems targeted to lyrical so maybe this should go to lyrical's branch? Unless you're working on that :-) 

Otherwise its much better

.github/deps.repos
u33549: I created a new PR for Lyrical. While porting this section to Rolling, I passed all the tests in the dev container, but the CI tests kept failing during the build, so I switched from Rolling to Lyrical. I’m currently fixing the CI issues, and after that, I’ll work on it a bit more and make some improvements for Rolling.

SteveMacenski:
Let me kno wwhen you want me to look again

u33549:
> Let me kno wwhen you want me to look again

Sure, if there are any improvements that need to be made to `open_nav_coverage` or in other repositories, you can contribute as well; I’m trying to fix or improve what I’ve seen.

u33549:
There is okey i guess. You can merge.

SteveMacenski (approved):

u33549:
@SteveMacenski I plan to start integrating the new features introduced with f2cv2 soon. Is there a different repository or any issues you’d recommend I look into, or would you prefer that I focus on this one?

SteveMacenski:
If you want to work on F2C integrations for Nav2, this would be the place! if you wanted to contribute to other Nav2 projects, then Nav2's home repo is probably the right place to look :-) 

Check out the issue tracker for any other projects w.r.t. this https://github.com/open-navigation/opennav_coverage/issues

u33549:
I’ll be working on the F2C integration for a while longer. When I looked at other projects, this seemed more reasonable in terms of cost-benefit, but CI/CD and local tests are taking too long. Isn’t there a way to use the packages we installed from source as binaries?

SteveMacenski:
Yes in Jazzy, Kilted, Humble, etc :- )

u33549:
I know they have it, but what about in Rolling?

SteveMacenski:
Nav2 does not and will not have rolling binaries anytime soon

u33549:
alright :(

SteveMacenski:
But local builds should be fast- just build once and you should have it for the duration of all your testing. You should never need to recompile Nav2

u33549:
I use devcontainer on my local machine so I don't have to keep changing the operating system; even though I cache the builds, whenever I make changes in Dockerfile, it has to rebuild from scratch. It's probably a skill issue on my part.

SteveMacenski:
Mount you workspace locally to a directory so the build artifacts persist between builds so its not internal to the container, at the very least

u33549:
I’d like to ask a quick question. When I submit a pr on GitHub, would you prefer that I submit it all at once or in parts? The integration is getting a bit large, so I was thinking of submitting a single pr once everything is finished, but then I thought about the review process and decided to ask you.
image


The photo is just a part of it; I'm still working on the rest.

SteveMacenski:
I’d break into logical chunks. 25x 5 line PRs is more annoying but 1x 2,000 line one is intractable. Use your best judgement. Keep them under ~3 features though

u33549:
All right, I'll try to balance things out.

u33549:
Do you want to see a few spoilers?
image
image
image

Decomposition is actively used within opennav_covrage and calculates the optimal scan path on its own

I guess I’m a bit of a chatterbox, but sharing things like this boosts my motivation to work—sorry about that.

SteveMacenski:
Ooo cool!

u33549:
While using it in my own project, I found some bugs, and I need to fix them across all distros. Do I need to open a separate PR for each one, or is there an easier way?

For now, I'll enable it for all distros so it can be fixed quickly, but it would be good for me to learn about any future ones, if there are any.

SteveMacenski:
Depends on how complicated the bugs are, but they should definitely be separated from exposing new features.

PR #112, opened 2026-07-02
Add task time, per-pose velocity/direction, and NSwathModified swath objective

u33549:
## Summary

First of a few PRs splitting a larger Fields2Cover (F2C) v2 enhancement branch
into reviewable chunks, as discussed. It bundles three small, self-contained
output/objective improvements plus one build fix. None of these depend on the
larger field-decomposition / TSP-routing work that will follow in later PRs.

## Changes

**1. Per-pose velocity and direction on `PathComponents`**
- Adds `float64[] velocities` (m/s) and `bool[] is_backward` to
  `PathComponents.msg`, filled in parallel with `nav_path.poses`.
- New `toNavPathMsg` overload populates the per-pose velocity/direction vectors
  alongside the path; filled only when `generate_path=true`, empty otherwise.
- Wired through both `coverage_server` and `row_coverage_server`.

**2. Task time on the action result + `discretizeSwath` refactor**
- `toNavPathMsg` now uses `Path::discretizeSwath(step_size)` instead of a
  hand-written linear-interpolation loop — only `SWATH` states are split,
  `TURN` states pass through unchanged.
- Adds `float64 task_time` to the `ComputeCoveragePath` action result,
  populated from `Path::getTaskTime()` with an `std::isfinite` guard for
  zero-velocity states.

**3. `NUMBER_MODIFIED` swath objective**
- Exposes Fields2Cover's `f2c::obj::NSwathModified` as a selectable swath
  objective via `NUMBER_MODIFIED`, handled in
  `SwathGenerator::createObjective/toString/toType`.

**4. Build fix: keep `-Wnull-dereference` non-fatal**
- GCC false-positives on inlined ROS 2 message `operator==/!=` (e.g.
  `coordinate__struct.hpp`) reached via `getFieldFromGoal()`, breaking
  `-Werror` builds on newer toolchains. The warning stays visible but is no
  longer fatal.

## Interface changes (all additive)

- `PathComponents.msg`: `+ float64[] velocities`, `+ bool[] is_backward`
- `ComputeCoveragePath.action` (result): `+ float64 task_time`
- `SwathMode.msg`: `objective` now also accepts `NUMBER_MODIFIED` (doc comment updated)

## Testing

- `test_utils`: `discretizeSwath` point count, `TURN` passthrough regression,
  `getTaskTime` normal/zero/empty; plus empty-path, forward-only,
  velocity/direction values, and vector-size-match cases.
- `test_swath`: `NUMBER_MODIFIED` round-trips through
  `toType`/`toString`/`createObjective`, plus a generation smoke test.

### DIFF
--- a/opennav_coverage/CMakeLists.txt
+++ b/opennav_coverage/CMakeLists.txt
@@ -22,7 +22,9 @@ find_package(GDAL REQUIRED)
 
 # potentially replace with nav2_common, nav2_package()
 set(CMAKE_CXX_STANDARD 20)
-add_compile_options(-Wall -Wextra -Wpedantic -Werror -Wdeprecated -fPIC -Wshadow -Wnull-dereference)
+# -Wno-error=null-dereference: GCC false-positives on inlined ROS2 message
+# operator==/!= (e.g. coordinate__struct.hpp) when included via getFieldFromGoal().
+add_compile_options(-Wall -Wextra -Wpedantic -Werror -Wdeprecated -fPIC -Wshadow -Wnull-dereference -Wno-error=null-dereference)
 
 include_directories(
   include
--- a/opennav_coverage/CMakeLists.txt
+++ b/opennav_coverage/CMakeLists.txt
@@ -22,8 +22,6 @@ find_package(GDAL REQUIRED)
 
 # potentially replace with nav2_common, nav2_package()
 set(CMAKE_CXX_STANDARD 20)
-# -Wno-error=null-dereference: GCC false-positives on inlined ROS2 message
-# operator==/!= (e.g. coordinate__struct.hpp) when included via getFieldFromGoal().
 add_compile_options(-Wall -Wextra -Wpedantic -Werror -Wdeprecated -fPIC -Wshadow -Wnull-dereference -Wno-error=null-dereference)
 
 include_directories(
--- a/opennav_coverage/include/opennav_coverage/utils.hpp
+++ b/opennav_coverage/include/opennav_coverage/utils.hpp
@@ -237,6 +237,44 @@ inline nav_msgs::msg::Path toNavPathMsg(
 }
 
 /**
+ * @brief Overload of toNavPathMsg that also populates per-pose velocity and direction vectors,
+ * parallel to the returned nav_path.poses. Vectors are cleared before filling.
+ */
+inline nav_msgs::msg::Path toNavPathMsg(
+  const Path & raw_path, const F2CField & field,
+  const std_msgs::msg::Header & header, const bool is_cartesian,
+  const float & pt_dist,
+  std::vector<double> & out_velocities,
+  std::vector<bool> & out_is_backward)
+{
+  nav_msgs::msg::Path msg;
+  msg.header = header;
+  out_velocities.clear();
+  out_is_backward.clear();
+
+  if (raw_path.size() == 0) {
+    return msg;
+  }
+
+  Path path = raw_path;
+  if (!is_cartesian) {
+    path = f2c::Transform::transformToPrevCRS(raw_path, field);
+  } else {
+    path.moveTo(field.getRefPoint());
+  }
+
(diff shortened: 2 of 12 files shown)


u33549:
The remaining 3 pr are on their way. I split them up this way to make them easier to review.

u33549:
A new ci/cd error ./ I'll fix it soon. 

image

It was working on my computer

opennav_coverage/CMakeLists.txt
```diff
@@ -22,7 +22,9 @@ find_package(GDAL REQUIRED)
 
 # potentially replace with nav2_common, nav2_package()
 set(CMAKE_CXX_STANDARD 20)
-add_compile_options(-Wall -Wextra -Wpedantic -Werror -Wdeprecated -fPIC -Wshadow -Wnull-dereference)
+# -Wno-error=null-dereference: GCC false-positives on inlined ROS2 message
+# operator==/!= (e.g. coordinate__struct.hpp) when included via getFieldFromGoal().
```
SteveMacenski: remove comments

opennav_coverage/include/opennav_coverage/utils.hpp
```diff
@@ -226,38 +225,50 @@ inline nav_msgs::msg::Path toNavPathMsg(
...
+  }
+
+  return msg;
+}
+
+/**
+ * @brief Overload of toNavPathMsg that also populates per-pose velocity and direction vectors,
+ * parallel to the returned nav_path.poses. Vectors are cleared before filling.
+ */
+inline nav_msgs::msg::Path toNavPathMsg(
```
SteveMacenski, line 211: Why not just modify the existing function and drop that vel/backward data if not useful? In general though if we got it now, we should expose it. I think you basically just use this always now, so we can just update the 1 function rather than having 2x

opennav_coverage/include/opennav_coverage/utils.hpp
```diff
@@ -226,37 +231,15 @@ inline nav_msgs::msg::Path toNavPathMsg(
...
-      msg.poses.push_back(toMsg(path[i]));
+  // discretizeSwath splits only SWATH states at step_size intervals; TURN states pass through
+  // unchanged. Do NOT use discretize() — that calls populate()+reduce(), which modifies TURNs.
+  path = path.discretizeSwath(static_cast<double>(pt_dist));
+
+  for (const auto & state : path) {
+    msg.poses.push_back(toMsg(state));
+    if (out_velocities) {out_velocities->push_back(state.velocity);}
+    if (out_is_backward) {
+      out_is_backward->push_back(state.dir == f2c::types::PathDirection::BACKWARD);
```
SteveMacenski, line 247: For out_velocities, out_is_backward and msg.poses: do a `vector.reserve(size)` operation so that we don't dynamically reallocate during population

opennav_coverage/include/opennav_coverage/utils.hpp
```diff
@@ -226,37 +231,21 @@ inline nav_msgs::msg::Path toNavPathMsg(
...
-        pose.pose.position.x += pt_dist * ux;
-        pose.pose.position.y += pt_dist * uy;
-        msg.poses.push_back(pose);
-        curr_dist += pt_dist;
-      }
-    } else {
-      // Turns are already dense paths
-      msg.poses.push_back(toMsg(path[i]));
+  // discretizeSwath splits only SWATH states at step_size intervals; TURN states pass through
+  // unchanged. Do NOT use discretize() — that calls populate()+reduce(), which modifies TURNs.
```
SteveMacenski:
```suggestion
  // discretizeSwath splits only SWATH states at step_size intervals
```

SteveMacenski (approved):

u33549:
From now on, I will try to keep the comments shorter. I develop it with Ai and leave notes while developing it so that we don't have any problems in another session. I will pay more attention to unnecessary detailed comments before creating PR.

PR #113, opened 2026-07-03
Add field decomposition (trapezoidal / boustrophedon) with HL_SWATH handling

u33549:
## Summary

Second PR in the series splitting the Fields2Cover (F2C) v2 enhancement branch
into reviewable chunks. Adds optional **field decomposition** — splitting a
non-convex field into sub-cells before swath/headland generation — plus the
`HL_SWATH` path handling that decomposed fields require. Builds on #112.

## Changes

**Field decomposition (core)**
- `DecompType` enum: `NONE`, `TRAPEZOIDAL`, `BOUSTROPHEDON`.
- `DecompGenerator`: wraps the Fields2Cover decomposition, selectable via mode
  and `split_angle`.
- Multi-cell overloads for the swath and headland generators, operating on the
  `F2CCells` produced by decomposition.
- Wired into `coverage_server` behind a single gate: the `generate_decomp` goal
  flag **or** a `default_generate_decomp` YAML parameter.
- BT node ports exposing decomposition control.

**HL_SWATH handling**
- Decomposition emits `HL_SWATH` states for inter-cell headland passes.
  `toCoveragePathMsg`/`toNavPathMsg` now treat `HL_SWATH` like `SWATH`, so
  decomposed fields no longer crash the server.

## Interface changes (additive)

- `DecompMode.msg` (new): `string mode` (`NONE`/`TRAPEZOIDAL`/`BOUSTROPHEDON`),
  `float64 split_angle` (radians; `0` = parallel to swath angle).
- `ComputeCoveragePath.action`: `+ bool generate_decomp`, `+ DecompMode decomp_mode`.
- `types.hpp`: `+ enum class DecompType`.

## Testing

- Decomposition + multi-cell swath/headland generation unit tests.
- `HL_SWATH` path-conversion test (velocity/direction fill with `HL_SWATH` states).

## Note

This adds the decomposition machinery and message/BT plumbing. The TSP route
planning that consumes the decomposed cells follows in the next PR of the series.

PR #117, opened 2026-07-06
Add TSP route ordering (Fields2Cover genRoute / OR-Tools)

u33549:
## Summary

Third PR in the series splitting the Fields2Cover (F2C) v2 enhancement branch.
Adds a **TSP-based route orderer** — ordering the unordered swaths via F2C's
`genRoute` (OR-Tools) — as a new `TSP` route mode, alongside the existing
`BOUSTROPHEDON` / `SNAKE` / `SPIRAL` / `CUSTOM` orderers. Builds on #112 and #113.

## Changes

**TSP route planning (core)**
- `RouteType::TSP` and `RouteGenerator::generateRouteTSP` (+ `resolveType`),
  wrapping the Fields2Cover `genRoute` route planner with its OR-Tools knobs.
- Wired into `coverage_server`: when `route_mode = TSP`, swaths are ordered by
  the TSP planner and the connecting path is planned over that route.
- BT node ports exposing the TSP knobs.
- Per-cell TSP + route stitching for multi-cell input: a single `genRoute` call
  builds an all-pairs path matrix that blows up memory, so the route is planned
  per cell and stitched.

## Interface changes (additive)

- `types.hpp`: `RouteType` enum `+ TSP`.
- `RouteMode.msg`: `mode` now also accepts `TSP`; new TSP knobs
  `tsp_redirect_swaths` (bool), `tsp_time_limit` (s), `tsp_search_for_optimum` (bool).

## Testing

- `test_route`: TSP route generation, route-type resolution, and the TSP knobs.

## Note

This PR adds standalone TSP ordering for a single (non-decomposed) field or cell.
A follow-up PR will route decomposed (multi-cell) fields with the TSP planner,
connecting cells through their shared headland rather than a direct shortcut.
Further PRs beyond that will continue splitting out the remaining enhancements.

### DIFF
--- a/opennav_coverage/include/opennav_coverage/path_generator.hpp
+++ b/opennav_coverage/include/opennav_coverage/path_generator.hpp
@@ -78,8 +78,8 @@ public:
     const Swaths & swaths, const opennav_coverage_msgs::msg::PathMode & settings);
 
   /**
-   * @brief Generate path from a TSP F2CRoute (headland connections included).
-   * @param route Route produced by generateRouteTSP
+   * @brief Generate path from an F2CRoute (headland connections included).
+   * @param route Route produced by RouteGenerator::generateRoute
    * @param settings PathMode for curve selection
    * @return Path complete path including headland connections
    */
--- a/opennav_coverage/include/opennav_coverage/path_generator.hpp
+++ b/opennav_coverage/include/opennav_coverage/path_generator.hpp
@@ -78,7 +78,7 @@ public:
     const Swaths & swaths, const opennav_coverage_msgs::msg::PathMode & settings);
 
   /**
-   * @brief Generate path from a TSP F2CRoute (K4 minimal B.2 slice).
+   * @brief Generate path from a TSP F2CRoute (headland connections included).
    * @param route Route produced by generateRouteTSP
    * @param settings PathMode for curve selection
    * @return Path complete path including headland connections
--- a/opennav_coverage/include/opennav_coverage/path_generator.hpp
+++ b/opennav_coverage/include/opennav_coverage/path_generator.hpp
@@ -69,15 +69,6 @@ public:
   }
 
   /**
-   * @brief Main method to generate path
-   * @param Swaths swaths to generate path from
-   * @param request Action request information
-   * @return Path complete path
-   */
-  Path generatePath(
-    const Swaths & swaths, const opennav_coverage_msgs::msg::PathMode & settings);
-
-  /**
    * @brief Generate path from an F2CRoute (headland connections included).
    * @param route Route produced by RouteGenerator::generateRoute
    * @param settings PathMode for curve selection
--- a/opennav_coverage/include/opennav_coverage/path_generator.hpp
+++ b/opennav_coverage/include/opennav_coverage/path_generator.hpp
@@ -15,7 +15,6 @@
 #ifndef OPENNAV_COVERAGE__PATH_GENERATOR_HPP_
 #define OPENNAV_COVERAGE__PATH_GENERATOR_HPP_
 
-#include <utility>
 #include <vector>
 #include <string>
 #include <memory>
@@ -104,12 +103,6 @@ protected:
    * @return Generator to use
    */
   TurningBasePtr createCurve(const PathType & type, const PathContinuityType & c_type);
-
-  /**
-   * @brief Resolve curve and turn_point_distance from PathMode (shared by both overloads)
-   */
-  std::pair<TurningBasePtr, float> resolveCurve(
-    const opennav_coverage_msgs::msg::PathMode & settings);
 
   /**
    * @brief Converts the path mode into a string for publication
--- a/opennav_coverage/include/opennav_coverage/path_generator.hpp
+++ b/opennav_coverage/include/opennav_coverage/path_generator.hpp
@@ -15,6 +15,7 @@
 #ifndef OPENNAV_COVERAGE__PATH_GENERATOR_HPP_
 #define OPENNAV_COVERAGE__PATH_GENERATOR_HPP_
 
+#include <utility>
 #include <vector>
 #include <string>
 #include <memory>
--- a/opennav_coverage/include/opennav_coverage/path_generator.hpp
+++ b/opennav_coverage/include/opennav_coverage/path_generator.hpp
@@ -77,6 +77,15 @@ public:
     const Swaths & swaths, const opennav_coverage_msgs::msg::PathMode & settings);
 
   /**
+   * @brief Generate path from a TSP F2CRoute (K4 minimal B.2 slice).
+   * @param route Route produced by generateRouteTSP
+   * @param settings PathMode for curve selection
+   * @return Path complete path including headland connections
+   */
+  Path generatePath(
+    const F2CRoute & route, const opennav_coverage_msgs::msg::PathMode & settings);
+
+  /**
    * @brief Sets the mode manually of the paths for dynamic parameters
    * @param mode String for mode to use
    */
@@ -103,6 +112,12 @@ protected:
    * @return Generator to use
    */
   TurningBasePtr createCurve(const PathType & type, const PathContinuityType & c_type);
+
+  /**
+   * @brief Resolve curve and turn_point_distance from PathMode (shared by both overloads)
+   */
+  std::pair<TurningBasePtr, float> resolveCurve(
+    const opennav_coverage_msgs::msg::PathMode & settings);
 
   /**
    * @brief Converts the path mode into a string for publication
--- a/opennav_coverage/include/opennav_coverage/route_generator.hpp
+++ b/opennav_coverage/include/opennav_coverage/route_generator.hpp
@@ -87,29 +87,19 @@ public:
   }
 
   /**
-   * @brief Main method to generate route
-   * @param Swaths swaths to generate route from
-   * @param request Action request information
-   * @return Swaths ordered swaths
-   */
-  Swaths generateRoute(
-    const Swaths & swaths, const opennav_coverage_msgs::msg::RouteMode & settings);
-
-  /**
-   * @brief TSP-based multi-cell route planner using OR-Tools.
-   *        Requires generate_path=true; throws if called without it.
-   * @param cells Field cells (headland-removed) for connection routing
+   * @brief Generate an ordered route for any mode (orderers or TSP).
+   * @param cells Travel cells whose borders route connections may follow
    * @param swaths_by_cells Per-cell swaths from generateSwathsByCells
-   * @param settings RouteMode containing tsp_* knobs
-   * @return F2CRoute with ordered swath groups and headland connections
+   * @param settings Action request information
+   * @return F2CRoute (ordered swath groups plus any connections)
    */
-  F2CRoute generateRouteTSP(
+  F2CRoute generateRoute(
     const F2CCells & cells,
     const F2CSwathsByCells & swaths_by_cells,
     const opennav_coverage_msgs::msg::RouteMode & settings);
 
   /**
-   * @brief Resolve mode string to RouteType (public wrapper for server branching)
+   * @brief Resolve mode string to RouteType, applying the default when unset
    */
   RouteType resolveType(const opennav_coverage_msgs::msg::RouteMode & settings)
   {
--- a/opennav_coverage/include/opennav_coverage/route_generator.hpp
+++ b/opennav_coverage/include/opennav_coverage/route_generator.hpp
@@ -99,15 +99,6 @@ public:
     const opennav_coverage_msgs::msg::RouteMode & settings);
 
   /**
-   * @brief Resolve mode string to RouteType, applying the default when unset
-   */
-  RouteType resolveType(const opennav_coverage_msgs::msg::RouteMode & settings)
-  {
-    RouteType t = toType(settings.mode);
-    return (t == RouteType::UNKNOWN) ? default_type_ : t;
-  }
-
-  /**
    * @brief Sets the mode manually of the Route for dynamic parameters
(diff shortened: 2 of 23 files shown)


SteveMacenski:
What does this new ordering actually do?

This also seems to add more duplication than the other feature additions

opennav_coverage/include/opennav_coverage/route_generator.hpp
```diff
@@ -65,6 +65,25 @@ class RouteGenerator
         "default_custom_order was not set! "
         "If using Custom Route mode, the custom order must be set per-request!");
     }
+
+    nav2::declare_parameter_if_not_declared(
+      node, "default_tsp_redirect_swaths", rclcpp::ParameterValue(true));
```
SteveMacenski, line 70: Is default on the right default?

u33549:
The existing route modes (BOUSTROPHEDON/SNAKE/SPIRAL/CUSTOM) all order swaths by a fixed sweep pattern — they assume a single-cell field and decide visit order from that pattern alone, without looking at actual connection distances between swaths.

TSP instead wraps Fields2Cover's genRoute (RoutePlannerBase, OR-Tools-based): it treats visiting all swaths as a traveling-salesman problem — computing the real connection cost between every pair of swath endpoints and solving for the order (and direction) that minimizes total travel/connection distance, rather than assuming a serpentine pattern holds.

Where this actually matters: on decomposed / multi-region fields, a fixed pattern order doesn't know about cell boundaries and can produce long backtracking connections between regions. TSP evaluates the real geometry and picks a shorter route instead of assuming pattern adjacency.

Knobs exposed on RouteMode:

tsp_redirect_swaths — allow the solver to reverse a swath's direction if that gives a shorter connection
tsp_time_limit — OR-Tools solver time budget (seconds)
tsp_search_for_optimum — enable guided local search for a better (but slower) solve
This PR adds TSP ordering for a single field/cell. Routing decomposed multi-cell fields with it is a follow-up (uses the same generateRouteTSP, per-cell, then stitches the routes together — the per-cell/stitching rationale is in route_generator.cpp).

SteveMacenski:
Should we find an architectural way to have this work as a `RouteGeneratorPtr` rather than embedding in `RouteGenerator` directly? This breaks the design pattern using the route generator ptr & I think causes this breakout of if/else in the server itself as well causing so much duplication.

u33549:
The blocker is at the F2C level: `RouteGeneratorPtr` is a `SingleCellSwathsOrderBase` (`genSortedSwaths(F2CSwaths) -> F2CSwaths`), whereas the TSP planner is `RoutePlannerBase` — a separate, unrelated base class with different inputs (`F2CCells` + `F2CSwathsByCells`) and a different output (`F2CRoute`). So it can't slot into the existing factory as-is, which is what forced the is_tsp branch into the server.

The fix I'd propose is to add a thin abstraction on our side that both implement and that returns a unified `F2CRoute`:


```
class RouteMethod {
 public:
  virtual F2CRoute plan(const F2CCells & travel_cells,
                        const F2CSwathsByCells & swaths_by_cells,
                        const RouteMode & settings) = 0;
};
using RouteGeneratorPtr = std::shared_ptr<RouteMethod>;
```
`SwathOrderMethod` wraps the existing `SingleCellSwathsOrderBase` orderers (flattens swaths, calls `genSortedSwaths`, wraps the result in a single-group `F2CRoute`).
`TspRouteMethod` wraps `RoutePlannerBase::genRoute.`
`createGenerator() `returns the right one, so the server just calls `route_gen_->plan(...)` with no i`s_tsp` branching.
Normalizing both paths to `F2CRoute` also removes the duplication that branch created — the second swath generation and the repeated result-filling collapse into one path.

Sound good?

SteveMacenski:
Sounds good to me

u33549:
i think this version is better

opennav_row_coverage/src/row_coverage_server.cpp
```diff
@@ -198,7 +198,15 @@ void RowCoverageServer::computeCoveragePath()
     header.frame_id = frame_id;
     Path path;
     if (goal->generate_route) {
-      Swaths route = route_gen_->generateRoute(swaths, goal->route_mode);
+      // Single-cell group; the field polygon provides travel borders for TSP
+      F2CCells cells;
+      cells.addGeometry(field);
+      F2CSwathsByCells swaths_by_cells;
```
SteveMacenski, line 204: It seems like uses of many APIs are being replaced by new types. Does it make sense to keep the old? This PR and others you have opened have done this, creating a new API with a new input/output type while leaving the old in place. 

I understood this to mean there were paths that those methods were still used; if not, remove them completely and replace with this new API if the new API is the only used

opennav_row_coverage/src/row_coverage_server.cpp
```diff
@@ -213,9 +221,16 @@ void RowCoverageServer::computeCoveragePath()
         const double task_time = path.getTaskTime();
         result->task_time = std::isfinite(task_time) ? task_time : 0.0;
       } else {
+        // Ordered swaths only (no connecting turns)
+        Swaths ordered;
+        for (const auto & group : route.getVectorSwaths()) {
+          for (const auto & s : group) {
+            ordered.emplace_back(s);
+          }
+        }
```
SteveMacenski: Why is `route` not already properly populated?

opennav_coverage/include/opennav_coverage/path_generator.hpp
```diff
@@ -76,6 +77,15 @@ class PathGenerator
...
     const Swaths & swaths, const opennav_coverage_msgs::msg::PathMode & settings);
 
+  /**
+   * @brief Generate path from an F2CRoute (headland connections included).
+   * @param route Route produced by RouteGenerator::generateRoute
+   * @param settings PathMode for curve selection
+   * @return Path complete path including headland connections
+   */
+  Path generatePath(
+    const F2CRoute & route, const opennav_coverage_msgs::msg::PathMode & settings);
```
SteveMacenski: This, for example

opennav_coverage/include/opennav_coverage/route_generator.hpp
u33549: it's default on F2C i didn't want to change it

opennav_row_coverage/src/row_coverage_server.cpp
u33549: I added a unit test to compare whether both APIs return the same result in terms of accuracy. However I will delete it.

opennav_row_coverage/src/row_coverage_server.cpp
u33549: `route` is fully populated, this loop isn't filling in missing data, it's just a type conversion. With `generate_path=false` only the ordered swaths go into the result; the route's connections are only meaningful in the full path output (`generate_path=true`), which consumes `route` directly.

SteveMacenski: Why isn't that already converted when populating `route`?

opennav_row_coverage/src/row_coverage_server.cpp
SteveMacenski: Good thinking to verify correctness; after you do so though you can remove any unused methods. I hate dead code, especially if its duplicated

opennav_row_coverage/src/row_coverage_server.cpp
u33549: You're right. I noticed what you said, too. 
I'm moving the conversion process out of the servers and into `toCoveragePathMsg` rn.

SteveMacenski:
LMK when I should look again

u33549:
@SteveMacenski, can you check the last commits?

opennav_coverage/src/path_generator.cpp
```diff
@@ -20,18 +20,15 @@
 namespace opennav_coverage
 {
 
-Path PathGenerator::generatePath(
-  const Swaths & swaths, const opennav_coverage_msgs::msg::PathMode & settings)
+std::pair<TurningBasePtr, float> PathGenerator::resolveCurve(
```
SteveMacenski: With this new refactor, is changing this to a new function required?

u33549: Not anymore

opennav_coverage/src/coverage_server.cpp
```diff
@@ -196,39 +196,37 @@ void CoverageServer::computeCoveragePath()
...
-      // Apply a separate headland to each sub-cell
-      F2CCells cells_no_headland = decomposed;
-      if (goal->generate_headland) {
-        cells_no_headland = headland_gen_->generateHeadlands(decomposed, goal->headland_mode);
-      }
-      swaths = swath_gen_->generateSwaths(cells_no_headland, goal->swath_mode);
+      cells = goal->generate_headland ?
+        headland_gen_->generateHeadlands(decomposed, goal->headland_mode) : decomposed;
     } else {
-      // (1) Optional: Remove headland from polygon field
```
SteveMacenski, line 212: You remove the (1) and (2) comments so now the code comments just start at (3)

opennav_coverage/test/CMakeLists.txt
```diff
@@ -63,8 +63,12 @@ target_link_libraries(test_path
 )
 
 # Test server
+# TIMEOUT raised from the 60s default: test_server runs 7 tests in one binary;
+# testDecompPath (~27s) + testTSPDecompHeadlandPath (per-cell genRoute ~18s +
+# generatePath) together exceed 60s, so ctest killed the binary mid-test.
 ament_add_gtest(test_server
```
SteveMacenski: remove ai generated comments that explains what IT did not what the CODE does

u33549:
A question unrelated to the topic how are you able to review things so carefully? Do you have a strategy?

SteveMacenski:
Otherwise LGTM 

> A question unrelated to the topic how are you able to review things so carefully? Do you have a strategy?

Years of practice and I'm sure I still miss some.

u33549:
I fixed the comment issues.


> Years of practice and I'm sure I still miss some.

That's really impressive. I spent nearly four years writing code in a different area, and my coding and code review skills had really improved, but since I started working on robotics and with the advancements in AI, I've become really rusty. I didn't really notice it in my own projects, but working with you really brought it home to me. I think I need to go back to the old-school way of developing.

opennav_coverage/src/coverage_server.cpp
```diff
@@ -196,39 +196,37 @@ void CoverageServer::computeCoveragePath()
...
 
-    // (3) Optional: Generate an ordered route through the unordered swaths
+    // (2) Optional: Generate an ordered route through the unordered swaths
     std_msgs::msg::Header header;
     header.stamp = now();
     header.frame_id = frame_id;
     Path path;
     if (goal->generate_route) {
-      Swaths route = route_gen_->generateRoute(swaths, goal->route_mode);
+      // (2) One call for every mode: orderers and TSP both return an F2CRoute.
```
SteveMacenski: Remove this line

opennav_coverage/src/coverage_server.cpp
```diff
@@ -196,39 +196,37 @@ void CoverageServer::computeCoveragePath()
...
+      F2CRoute route = route_gen_->generateRoute(cells, swaths_by_cells, goal->route_mode);
+      if (route.isEmpty()) {
+        throw CoverageException("Route planner returned an empty route.");
+      }
 
-      // (4) Optional: Generate connection turns between ordered swaths
-      // Converts UTM back to GPS, if necessary, for action returns
       if (goal->generate_path) {
+        // (3) Plan connecting turns / headland connections between ordered swaths
```
SteveMacenski: Add the original comment back. Make that 3 now I guess. Remove the line you added

SteveMacenski:
I feel like we're playing wack-a-mole with comments. These are important ordered clarifications of the workload of the code. Keep _all_ of my comments as they exist aligned with their new locations they exist at. 

Remove:

```
// Cells + per-cell swaths, generated once and reused by the route planner.
```

u33549:
I'm really sorry for any misunderstandings. I'm asking about the changes I made here to make sure I understood correctly before pushing them.

```diff
     Field field_no_headland = field;
-    // Cells + per-cell swaths, generated once and reused by the route planner.
     F2CCells cells;
     if (do_decomp) {
       F2CCells raw_cells;
       raw_cells.addGeometry(field);
       F2CCells decomposed = decomp_gen_->decompose(raw_cells, goal->decomp_mode);
+
+      // Apply a separate headland to each sub-cell
       cells = goal->generate_headland ?
         headland_gen_->generateHeadlands(decomposed, goal->headland_mode) : decomposed;
     } else {
+      // (1) Optional: Remove headland from polygon field
       if (goal->generate_headland) {
         field_no_headland = headland_gen_->generateHeadlands(field, goal->headland_mode);
       }
       cells.addGeometry(field_no_headland);
     }
+
+    // (2) Generate swaths to cover polygon field, including internal voids
     F2CSwathsByCells swaths_by_cells = swath_gen_->generateSwathsByCells(cells, goal->swath_mode);
     Swaths swaths = swaths_by_cells.flatten();
 
-    // (2) Optional: Generate an ordered route through the unordered swaths
+    // (3) Optional: Generate an ordered route through the unordered swaths
     ...
-      // (3) Optional: Generate connection turns between ordered swaths
+      // (4) Optional: Generate connection turns between ordered swaths
       // Converts UTM back to GPS, if necessary, for action returns
```

Is this what you meant?

SteveMacenski:
Yes :-) As long as the comments are now colocated with the changed location of the key API calls to do each step

Edit: whoops, wrong button

u33549:
I hope I understood it correctly and made the right corrections.

SteveMacenski (approved):

PR #118, opened 2026-07-08
Route decomposed fields along headlands

u33549:
## Summary

Completes decomposition-with-route support: routes decomposed (multi-cell)
fields along their headlands using the F2C tutorial 7.2 ordering, and fixes a
multi-cell headland crash. Builds on #112, #113, #117.

## Changes

**Route decomposed fields along headlands (tutorial 7.2)**
- Previously a decomposed field applied the headland per sub-cell first, leaving
  each cell a disconnected island; per-cell TSP then stitched routes with
  straight-line bridges that cut across the field. Non-TSP modes flattened the
  multi-cell swaths into a single-cell orderer, fanning connections across the field.
- Now: shrink the whole field to a travel ring, decompose it, then shrink each
  sub-cell again for the swath area. `genRoute` is fed the border-sharing travel
  cells so inter-cell connections follow the headlands.
- `generateRouteTSP` takes separate `travel_cells` and `swath_cells`: small fields
  use a single global `genRoute`; large fields (swath count above a cap) fall back
  to per-cell TSP to avoid the O(N²) all-pairs OOM.
- Non-TSP route modes combined with decomposition are rejected with a clear
  `INVALID_MODE_SET` error.
- Decomposed cells and swaths are reused across the swath and TSP branches.

**Fix multi-cell headland crash on collapsed sub-cells**
- The multi-cell `generateHeadlands(F2CCells)` overload applied the headland to
  each sub-cell through the single-`Field` overload (ending in `.getGeometry(0)`).
  `ConstHL` is a negative buffer, so a sub-cell narrower than ~2×width collapses to
  empty and `getGeometry(0)` threw "Geometry does not contain point 0" on
  decomposed fields.
- Buffer each sub-cell directly, skip cells that collapse to empty, add every
  resulting polygon (a negative buffer may split one cell into several), and throw
  a clear `CoverageException` if every sub-cell collapses. Generator/width
  resolution extracted into a shared `resolveGenerator()` helper.

### DIFF
--- a/opennav_coverage/include/opennav_coverage/headland_generator.hpp
+++ b/opennav_coverage/include/opennav_coverage/headland_generator.hpp
@@ -87,6 +87,16 @@ public:
 
 protected:
   /**
+   * @brief Resolves the generator and width to use for a request, falling back
+   *        to the configured defaults when the request leaves the mode UNKNOWN.
+   * @param settings Action request information
+   * @param width Output: the headland width to apply
+   * @return Generator to use (never null; throws CoverageException otherwise)
+   */
+  HeadlandGeneratorPtr resolveGenerator(
+    const opennav_coverage_msgs::msg::HeadlandMode & settings, double & width);
+
+  /**
    * @brief Creates generator pointer of a requested type
    * @param type Headland generator type to create
    * @return Generator to use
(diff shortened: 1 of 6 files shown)


u33549:
In the next PR, or once everything is complete, can I update `opennav_coverage_demo` to showcase the new system? 

Edit: Maybe a balloon shape could be used in the demo.

u33549:
upsy wrong commits

PR #119, opened 2026-07-08
Route decomposed fields along headlands

u33549:
## Summary

Completes decomposition-with-route support: routes decomposed (multi-cell)
fields along their headlands so inter-cell connections follow the shared
headland, and fixes a multi-cell headland crash. Builds on #112, #113, #117.

## Changes

**Route decomposed fields along headlands**
- Previously a decomposed field applied the headland per sub-cell first, leaving
  each cell a disconnected island; per-cell TSP then stitched routes with
  straight-line bridges that cut across the field. Non-TSP modes flattened the
  multi-cell swaths into a single-cell orderer, fanning connections across the field.
- Now: shrink the whole field to a travel ring, decompose it, then shrink each
  sub-cell again for the swath area. `genRoute` is fed the border-sharing travel
  cells so inter-cell connections follow the headlands.
- `generateRouteTSP` takes separate `travel_cells` and `swath_cells`: small fields
  use a single global `genRoute`; large fields (swath count above a cap) fall back
  to per-cell TSP to avoid the O(N²) all-pairs OOM.
- Non-TSP route modes combined with decomposition are rejected with a clear
  `INVALID_MODE_SET` error.
- Decomposed cells and swaths are reused across the swath and TSP branches.

**Fix multi-cell headland crash on collapsed sub-cells**
- The multi-cell `generateHeadlands(F2CCells)` overload applied the headland to
  each sub-cell through the single-`Field` overload (ending in `.getGeometry(0)`).
  `ConstHL` is a negative buffer, so a sub-cell narrower than ~2×width collapses to
  empty and `getGeometry(0)` threw "Geometry does not contain point 0" on
  decomposed fields.
- Buffer each sub-cell directly, skip cells that collapse to empty, add every
  resulting polygon (a negative buffer may split one cell into several), and throw
  a clear `CoverageException` if every sub-cell collapses. Generator/width
  resolution extracted into a shared `resolveGenerator()` helper.

### DIFF
--- a/opennav_coverage/include/opennav_coverage/headland_generator.hpp
+++ b/opennav_coverage/include/opennav_coverage/headland_generator.hpp
@@ -87,6 +87,16 @@ public:
 
 protected:
   /**
+   * @brief Resolves the generator and width to use for a request, falling back
+   *        to the configured defaults when the request leaves the mode UNKNOWN.
+   * @param settings Action request information
+   * @param width Output: the headland width to apply
+   * @return Generator to use (never null; throws CoverageException otherwise)
+   */
+  HeadlandGeneratorPtr resolveGenerator(
+    const opennav_coverage_msgs::msg::HeadlandMode & settings, double & width);
+
+  /**
    * @brief Creates generator pointer of a requested type
    * @param type Headland generator type to create
    * @return Generator to use
--- a/opennav_coverage/src/coverage_server.cpp
+++ b/opennav_coverage/src/coverage_server.cpp
@@ -192,7 +192,7 @@ void CoverageServer::computeCoveragePath()
       "Generating coverage path in %s frame for zone with %zu outer nodes and %zu inner polygons.",
       frame_id.c_str(), field.getGeometry(0).size(), field.size() - 1);
 
-    // (1) Optional: decompose non-convex field, then remove headland, then generate swaths
+    // (1) Build the cells to cover: remove the headland, optionally decomposing first
     const bool do_decomp = goal->generate_decomp || default_generate_decomp_;
 
     Field field_no_headland = field;
@@ -201,8 +201,9 @@ void CoverageServer::computeCoveragePath()
     F2CCells route_cells;
     F2CCells swath_cells;
     if (do_decomp) {
-      // Shrink for the travel ring first, decompose it, then shrink each sub-cell
-      // again for its swath area, so connections follow the shared headland.
+      // Headland-first: remove the headland from the whole field, decompose that into
+      // border-sharing cells (the travel graph), then remove the headland again per
+      // cell for the swaths, so connections run along the shared headland.
       Field travel_ring = field;
       if (goal->generate_headland) {
         travel_ring = headland_gen_->generateHeadlands(field, goal->headland_mode);
@@ -214,6 +215,7 @@ void CoverageServer::computeCoveragePath()
       swath_cells = goal->generate_headland ?
         headland_gen_->generateHeadlands(route_cells, goal->headland_mode) : route_cells;
     } else {
+      // No decomposition: remove the headland from the single field cell
       if (goal->generate_headland) {
         field_no_headland = headland_gen_->generateHeadlands(field, goal->headland_mode);
       }
--- a/opennav_coverage/src/coverage_server.cpp
+++ b/opennav_coverage/src/coverage_server.cpp
@@ -339,6 +339,8 @@ CoverageServer::dynamicParametersCallback(std::vector<rclcpp::Parameter> paramet
         route_gen_->setSpiralN(parameter.as_int());
       } else if (name == "default_tsp_time_limit") {
         route_gen_->setTspTimeLimit(parameter.as_int());
+      } else if (name == "default_max_swaths_for_global_route") {
+        route_gen_->setMaxSwathsForGlobalRoute(parameter.as_int());
       }
     } else if (type == ParameterType::PARAMETER_INTEGER_ARRAY) {
       if (name == "default_custom_order") {
--- a/opennav_coverage/src/coverage_server.cpp
+++ b/opennav_coverage/src/coverage_server.cpp
@@ -196,25 +196,34 @@ void CoverageServer::computeCoveragePath()
     const bool do_decomp = goal->generate_decomp || default_generate_decomp_;
 
     Field field_no_headland = field;
-    F2CCells cells;
+    // route_cells: travel graph for the route planner. swath_cells: cells swaths
+    // are generated from. Differ only when decomposing with a headland.
+    F2CCells route_cells;
+    F2CCells swath_cells;
     if (do_decomp) {
-      F2CCells raw_cells;
-      raw_cells.addGeometry(field);
-      F2CCells decomposed = decomp_gen_->decompose(raw_cells, goal->decomp_mode);
-
-      // Apply a separate headland to each sub-cell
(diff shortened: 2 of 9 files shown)


u33549:
In the next PR, or once everything is complete, can I update opennav_coverage_demo to showcase the new system?

I tried symbols like a balloon or a navigation icon, but they didn't look very good. I'll probably try a simpler letter or shape.

u33549:
It also works well with simple shapes, but as the shape becomes more complex, for example, as gaps are added or the concavity increases, it can sometimes produce different results. Once the integration is complete, I’ll try to set aside some time specifically for making improvements.

opennav_coverage/include/opennav_coverage/headland_generator.hpp
```diff
@@ -86,6 +86,16 @@ class HeadlandGenerator
...
 
 protected:
+  /**
+   * @brief Resolves the generator and width to use for a request, falling back
+   *        to the configured defaults when the request leaves the mode UNKNOWN.
+   * @param settings Action request information
+   * @param width Output: the headland width to apply
+   * @return Generator to use (never null; throws CoverageException otherwise)
+   */
+  HeadlandGeneratorPtr resolveGenerator(
```
SteveMacenski, line 96: Same question as the other PR: Do we need the 2x instances of `generateHeadlands` still with the new changes to require the new method?

opennav_coverage/src/coverage_server.cpp
```diff
@@ -196,25 +196,34 @@ void CoverageServer::computeCoveragePath()
...
+        travel_ring = headland_gen_->generateHeadlands(field, goal->headland_mode);
+        field_no_headland = travel_ring;
+      }
+      F2CCells travel_ring_cells;
+      travel_ring_cells.addGeometry(travel_ring);
+      route_cells = decomp_gen_->decompose(travel_ring_cells, goal->decomp_mode);
+      swath_cells = goal->generate_headland ?
+        headland_gen_->generateHeadlands(route_cells, goal->headland_mode) : route_cells;
     } else {
-      // (1) Optional: Remove headland from polygon field
```
SteveMacenski, line 209: 👀 We need a step (1) then somewhere for removing headlands or whatever the new "step 1" is

opennav_coverage/src/route_method.cpp
```diff
@@ -19,11 +19,23 @@
 namespace opennav_coverage
 {
 
+// Above this many total swaths, TspRouteMethod falls back to per-cell TSP to avoid
+// F2C's O(N^2) all-pairs path matrix that OOMs on large decompositions.
+static constexpr size_t kMaxSwathsForGlobalRoute = 300;
```
SteveMacenski: How did you determine this value and what computer did you test it against that caused memory issues?

SteveMacenski:
> Can I update opennav_coverage_demo to showcase the new system?

Sure! That'd be great. Maybe mention some of the new features too in the readme that this adds if there's a logical place for it

opennav_coverage/src/coverage_server.cpp
u33549: I had developed this PR before merging the previous one. It slipped my mind while I was revising it, so I'm fixing it now.

opennav_coverage/include/opennav_coverage/headland_generator.hpp
u33549: Unlike #117 these both have live callers:` generateHeadlands(Field) `for the whole-field headland (non-decomp + the travel ring), and `generateHeadlands(F2CCells)` for the per-sub-cell headland in the decomposition path (where the collapse fix lives). Different return types and semantics, so both are needed.

opennav_coverage/src/route_method.cpp
u33549: Honestly the value is approximate — I hit an actual out-of-memory crash (blue-screen) while testing a very large field on my own machine, so I set a conservative cap below where that happened rather than a precisely-tuned number. The root cause is algorithmic: F2C v2.0.0's `RoutePlannerBase` materializes the all-pairs shortest-path matrix (~N² with N ≈ 4× total swaths), so it blows up on large decompositions regardless of the machine. However, it might be helpful to make this parametric.

This is fixed upstream in F2C commit [0ce7fcc](https://github.com/Fields2Cover/Fields2Cover/commit/0ce7fcce82f3f79f773699448769bcd679c739bc), but that landed on main after the v2.0.0 release and we depend on the released v2, so we can't pull it in yet.

SteveMacenski: OK. How about stating a warning in the docs instead of setting an arbitrary limit?

u33549: But wouldn’t that eliminate its usability in these large areas? Wouldn’t it make more sense to convert it to a parameter and add a warning?

Also, the per-cell fallback isn't only a theoretical guard, it's what keeps large decomposed fields returning a valid (if less-optimal) route instead of crashing. Concretely, `testTSPDecompHeadlandPath` already exercises it: `test_field.xml` has enough swaths to cross the threshold, so removing the fallback would send that test through a single genRoute over the whole swath set ,exactly the case that risks the OOM/timeout in CI.

opennav_coverage/src/headland_generator.cpp
```diff
@@ -37,21 +36,48 @@ Field HeadlandGenerator::generateHeadlands(
...
+  // Apply the headland to each sub-cell independently: decomposed cells share
+  // borders, so buffering the whole F2CCells at once would only shrink the outer
+  // boundary and drop the inter-cell headlands.
   F2CCells result;
   for (size_t i = 0; i < cells.size(); ++i) {
-    result.addGeometry(generateHeadlands(cells.getGeometry(i), settings));
+    // A sub-cell narrower than ~2*width collapses to an empty geometry under the
+    // inward buffer; F2C then returns zero cells and getGeometry(0) would throw
+    // "Geometry does not contain point 0". Skip such cells. The buffer can also
+    // split one cell into several polygons, so add every resulting geometry.
```
SteveMacenski: Make this less verbose, but there is clarity that the comment offers so you should keep it

opennav_coverage/src/route_method.cpp
```diff
@@ -56,10 +68,22 @@ F2CRoute TspRouteMethod::plan(
...
+  // square of swath count and can run out of memory on large decompositions.
+  size_t total_swaths = 0;
+  for (size_t i = 0; i < swaths_by_cells.size(); ++i) {
+    total_swaths += swaths_by_cells.at(i).size();
+  }
+  if (total_swaths <= kMaxSwathsForGlobalRoute) {
+    f2c::rp::RoutePlannerBase rp;
+    return rp.genRoute(
+      cells, swaths_by_cells,
+      /*show_log=*/false, d_tol, redirect_swaths, time_limit, search_for_optimum);
```
SteveMacenski: show log? Remove the comment

opennav_coverage/src/route_method.cpp
SteveMacenski: OK - parameterize then :-)

SteveMacenski (approved):

u33549:
There are probably just a few PRs left before this development is complete. After that, I’ll entegrate to other distros. I also asked the F2C developer to add commit [0ce7fcc](https://github.com/Fields2Cover/Fields2Cover/commit/0ce7fcce82f3f79f773699448769bcd679c739bc) to v2.0.0

PR #120, opened 2026-07-11
Add TSP start pose, optional path reduce, and Rolling demo launch fix

u33549:
## Summary

Three small, independent improvements to the coverage server. Each change is opt-in or behavior-preserving by default, so this PR is backwards compatible.

1. Optional TSP start/return pose
2. Optional path point reduction
3. Rolling demo launch fix

## 1. TSP start pose

Adds two goal fields to `ComputeCoveragePath`:

```
bool use_start_pose False
opennav_coverage_msgs/Coordinate start_pose
```

When `use_start_pose` is true and `route_mode` is TSP, the route starts and returns at `start_pose` (via `RoutePlannerBase::setStartAndEndPoint`). The point is given in the same frame as `polygons` and converted into the field's local frame, handling both the GPS/UTM and cartesian paths. For non-TSP route modes the start pose is ignored with a warning. The point is also exposed through the `ComputeCoveragePath` behavior-tree node ports.

## 2. Optional path reduce

Adds two dynamic parameters:

- `default_reduce_path` (bool, default `false`)
- `default_reduce_min_dist` (double, default `0.1`)

When enabled, `generatePath` runs F2C's `Path::reduce` to drop near-duplicate points (e.g. in turns). Default off, so the path output is unchanged unless explicitly requested.

## 3. Rolling demo launch fix

- `static_transform_publisher` doesn't accept positional arguments on Rolling; switched to named `--x/--y/.../--frame-id` flags in `coverage_demo_launch.py` and `row_coverage_demo_launch.py`.
- `bt_navigator` renamed `error_code_names` to `error_code_name_prefixes`; the old key threw during `CoverageNavigator` configure and segfaulted the component container, so the demo never came up.

### DIFF
--- a/opennav_coverage/include/opennav_coverage/utils.hpp
+++ b/opennav_coverage/include/opennav_coverage/utils.hpp
@@ -339,6 +339,22 @@ inline F2CField getFieldFromGoal(
 }
 
 /**
+ * @brief Converts a goal-frame point into the frame the field's swaths/cells live in
+ * @param pt Point in the goal frame (cartesian or GPS)
+ * @param field Field already transformed to its working frame
+ * @param is_cartesian Whether the goal coordinates are cartesian
+ * @return Point in the field's local frame
+ */
+inline F2CPoint toFieldFrame(const F2CPoint & pt, const F2CField & field, bool is_cartesian)
+{
+  // F2CField always stores geometry with its ref point subtracted, so the start point needs
+  // the same ref point removed (and, for GPS, projecting to UTM first).
+  F2CPoint abs = is_cartesian ?
+    pt : f2c::Transform::transform(pt, field.getPrevCRS(), field.getCRS());
+  return abs - field.getRefPoint();
+}
+
+/**
  * @brief Converts a string to uppercase
  * @param string String to change to uppercase
  */
--- a/opennav_coverage/src/route_method.cpp
+++ b/opennav_coverage/src/route_method.cpp
@@ -25,8 +25,6 @@ F2CRoute SwathOrderMethod::plan(
   const opennav_coverage_msgs::msg::RouteMode & settings,
   const std::optional<F2CPoint> & /*start_end_point*/)
 {
-  // start_end_point is a TSP-only concept; the generator warns the user
-
   // The orderers assume a single cell; multi-cell (decomposed) input breaks their
   // ordering, so only TSP handles it. Reject rather than produce a bad route.
   if (cells.size() > 1) {
--- a/opennav_coverage/src/route_method.cpp
+++ b/opennav_coverage/src/route_method.cpp
@@ -22,8 +22,11 @@ namespace opennav_coverage
 F2CRoute SwathOrderMethod::plan(
   const F2CCells & cells,
   const F2CSwathsByCells & swaths_by_cells,
-  const opennav_coverage_msgs::msg::RouteMode & settings)
+  const opennav_coverage_msgs::msg::RouteMode & settings,
+  const std::optional<F2CPoint> & start_end)
 {
+  (void)start_end;  // start/end is a TSP-only concept; the generator warns the user
+
   // The orderers assume a single cell; multi-cell (decomposed) input breaks their
   // ordering, so only TSP handles it. Reject rather than produce a bad route.
   if (cells.size() > 1) {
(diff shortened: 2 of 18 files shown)


opennav_coverage/include/opennav_coverage/utils.hpp
```diff
@@ -338,6 +338,22 @@ inline F2CField getFieldFromGoal(
...
+ */
+inline F2CPoint toFieldFrame(const F2CPoint & pt, const F2CField & field, bool is_cartesian)
+{
+  // F2CField always stores geometry with its ref point subtracted, so the start point needs
+  // the same ref point removed (and, for GPS, projecting to UTM first).
+  F2CPoint abs = is_cartesian ?
+    pt : f2c::Transform::transform(pt, field.getPrevCRS(), field.getCRS());
+  return abs - field.getRefPoint();
+}
+
```
SteveMacenski, line 356: If this is now a standalone util, are there places this workload is being used we should replace to use this util?

opennav_coverage/src/route_method.cpp
```diff
@@ -22,8 +22,11 @@ namespace opennav_coverage
 F2CRoute SwathOrderMethod::plan(
   const F2CCells & cells,
   const F2CSwathsByCells & swaths_by_cells,
-  const opennav_coverage_msgs::msg::RouteMode & settings)
+  const opennav_coverage_msgs::msg::RouteMode & settings,
+  const std::optional<F2CPoint> & start_end)
 {
+  (void)start_end;  // start/end is a TSP-only concept; the generator warns the user
```
SteveMacenski: comment out in the method then `/*start_end*/` don't `(void)` op it. 

Also `start_end` is not a really good... variable name.

SteveMacenski (commented):
I get the other features, but what's with `use_start_pose`? Can you describe the need/usecase?

u33549:
> I get the other features, but what's with `use_start_pose`? Can you describe the need/usecase?

With this feature, you can add a starting point to your coverage plan. When you add it, the plan starts and ends at that point.

For example the field swath start with (15,20) but the vehicle is waiting at (0.0). You can add (0,0) as a start point.

Edit: This feature is one of the features that came with f2cv2, in a way, I just added api.

opennav_coverage/src/route_method.cpp
u33549: Alr, I haven't had my computer with me for a few days, I'll fix it as soon as possible.

opennav_coverage/include/opennav_coverage/utils.hpp
u33549: I need to check, if there is, I will update it.

SteveMacenski:
Got it, that actually might be nice to update to use in our demo too

u33549:
The next pr is ready, I updated some things about the field coverage there. I have increased processing load and accuracy. I asked Mier about what I did and he said it wasn't a bad idea.
image_20260714_022207c4a1eac3-c6e4-463d-bdd8-51e8b3fb1f15
There is a spoiler.

opennav_coverage/src/route_method.cpp
u33549: I renamed to `start_end_point` because f2cv2 uses the names `r_start_end` / `setStartAndEndPoint`.

opennav_coverage/include/opennav_coverage/utils.hpp
u33549: i caugh two at opennav_row_coverage transformRowsWithRef, removeRowsRefPoint but they work a bit different so I recommend not changing it.

opennav_coverage/src/route_method.cpp
```diff
@@ -22,8 +22,11 @@ namespace opennav_coverage
 F2CRoute SwathOrderMethod::plan(
   const F2CCells & cells,
   const F2CSwathsByCells & swaths_by_cells,
-  const opennav_coverage_msgs::msg::RouteMode & settings)
+  const opennav_coverage_msgs::msg::RouteMode & settings,
+  const std::optional<F2CPoint> & /*start_end_point*/)
 {
+  // start_end_point is a TSP-only concept; the generator warns the user
```
SteveMacenski: Remove comment; this should be in doxygen only

SteveMacenski (commented):
Otherwise LGTM

SteveMacenski (approved):

PR #127, opened 2026-07-15
Multi-cell TSP routing, route corridors, and optional headland driving for decomposed fields

u33549:
## Summary

Every change is opt-in or behavior-preserving by default, so a request that doesn't
decompose the field (or leaves the new fields unset) plans exactly as before.

1. Multi-cell TSP routing (per-cell solve + stitch)
2. Route corridors only between adjacent cells (`route_headland_width`)
3. Path follows route connections when they detour
4. Optional headland driving (`generate_headland_swaths`)
5. Non-TSP + decomposition rejected early; dynamic turn params
6. test_server teardown fix

## 1. Multi-cell TSP routing

With decomposition, the field becomes several cells and a single global
`genRoute` both interleaves swaths across cells and builds F2C's all-pairs path
matrix, which grows with the square of the swath count and can run out of memory
on large fields.

Instead:

- **Single cell:** one `genRoute` call, which still honors the start/end point
  exactly.
- **Multi-cell:** each cell is solved on its own, then the per-cell routes are
  stitched in nearest-neighbor order. Each cell is entered from whichever end is
  cheaper (with a heading-mismatch penalty so the entry doesn't force an
  S-maneuver), and the bridge between two cells follows the shared border graph
  of that travel-cell pair, so it detours around voids instead of cutting a
  straight line through them.

## 2. Route corridors between adjacent cells

`route_headland_width` (default `0.0` = operation width) carves a travel
corridor **only** on borders shared by two adjacent decomposed cells. Edges that
already have a headland are left untouched — a corridor is only ever added
between two cells, never on the field boundary.

## 3. Path follows route connections

F2C's path planner drops connections whose endpoints are near a swath end, which
erased the inter-cell / around-void travel from `nav_path`. The path generator
now follows those connections (as `HL_SWATH` sections) so the detour actually
appears in the output path.

## 4. Optional headland driving

Adds goal fields to optionally cover the headland ring ("drive the headland"):

```
bool generate_headland_swaths False
bool headland_first          False   # drive headland before inner coverage (else after)
bool headland_full_coverage  False   # concentric band sweep (else a single perimeter pass)
```

All default false, so headland coverage is off unless requested.
`default_generate_headland_swaths` is also exposed as a parameter.

## 5. Rejections and dynamic params

- Non-TSP route modes (BOUSTROPHEDON/SNAKE/SPIRAL/CUSTOM) don't have a
  cross-cell ordering, so combining them with decomposition is rejected early
  with a clear error instead of producing a bad route.
- `min_turning_radius`, `linear_curv_change`, and
  `default_generate_headland_swaths` are now dynamic parameters.

## 6. test_server teardown fix

Tearing the server down while the action worker was still running could corrupt
the heap and segfault the test. Teardown now waits for the worker to finish.

### DIFF
--- a/opennav_coverage/include/opennav_coverage/coverage_server.hpp
+++ b/opennav_coverage/include/opennav_coverage/coverage_server.hpp
@@ -132,6 +132,7 @@ protected:
   std::unique_ptr<Visualizer> visualizer_;
   bool cartesian_frame_;
   bool default_generate_decomp_{false};
+  bool default_generate_headland_swaths_{false};
 };
 
 }  // namespace opennav_coverage
--- a/opennav_coverage/include/opennav_coverage/headland_generator.hpp
+++ b/opennav_coverage/include/opennav_coverage/headland_generator.hpp
@@ -75,6 +75,16 @@ public:
     const F2CCells & cells, const opennav_coverage_msgs::msg::HeadlandMode & settings);
 
   /**
+   * @brief Carve a route corridor only on the borders cells actually share,
+   *        leaving edges facing the outer boundary or a void untouched.
+   * @param cells Decomposed sub-cells (share internal borders)
+   * @param route_width Corridor width carved along each shared border
+   * @return Swath cells carved only on their internal borders
+   */
+  F2CCells generateHeadlandsBetweenCells(
+    const F2CCells & cells, double route_width);
+
+  /**
    * @brief Generate concentric rings that fully sweep the headland band.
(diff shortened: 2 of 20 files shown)


u33549:
I think its the last pr for rolling after that I will control everything again but my first goal is developing other distros to the same level as rolling. After then we can prepare a guide.

SteveMacenski:
This one's a bit too big, can you break it up into 2-3 PRs?

PR #131, opened 2026-07-16
Field decomposition: route corridors between adjacent cells and optional headland driving

u33549:
## Summary

Route corridors between adjacent cells and optional headland driving for
**decomposed** fields. Every change is opt-in or behavior-preserving by default,
so a request that doesn't decompose the field (or leaves the new fields unset)
plans exactly as before.

1. Route corridors only between adjacent cells (`route_headland_width`)
2. Optional headland driving (`generate_headland_swaths`)
3. Non-TSP + decomposition rejected early; dynamic turn params

## 1. Route corridors between adjacent cells

`route_headland_width` (default `0.0` = operation width) carves a travel
corridor **only** on borders shared by two adjacent decomposed cells. Edges that
already have a headland are left untouched — a corridor is only ever added
between two cells, never on the field boundary.

## 2. Optional headland driving

Adds goal fields to optionally cover the headland ring ("drive the headland"):

```
bool generate_headland_swaths False
bool headland_first          False   # drive headland before inner coverage (else after)
bool headland_full_coverage  False   # concentric band sweep (else a single perimeter pass)
```

All default false, so headland coverage is off unless requested.
`default_generate_headland_swaths` is also exposed as a parameter.

## 3. Rejections and dynamic params

- Non-TSP route modes (BOUSTROPHEDON/SNAKE/SPIRAL/CUSTOM) don't have a
  cross-cell ordering, so combining them with decomposition is rejected early
  with a clear error instead of producing a bad route.
- `min_turning_radius`, `linear_curv_change`, and
  `default_generate_headland_swaths` are now dynamic parameters.

## Also included

- The `headland_swaths` RViz marker now publishes `DELETEALL` when a goal drives
  no headland pass, so a previous goal's loop doesn't linger.
- test_server teardown now waits for the action worker to finish (fixes a
  heap-corruption/segfault in the tests).

### DIFF
--- a/opennav_coverage/src/coverage_server.cpp
+++ b/opennav_coverage/src/coverage_server.cpp
@@ -206,9 +206,8 @@ void CoverageServer::computeCoveragePath()
     F2CCells route_cells;
     F2CCells swath_cells;
     if (do_decomp) {
-      // Headland-first: remove the headland from the whole field, decompose that into
-      // border-sharing cells (the travel graph), then remove the headland again per
-      // cell for the swaths, so connections run along the shared headland.
+      // Headland-first: remove the headland once from the whole field, then
+      // decompose that into border-sharing cells (the travel graph).
       Field travel_ring = field;
       if (goal->generate_headland) {
         travel_ring = headland_gen_->generateHeadlands(field, goal->headland_mode);
@@ -217,8 +216,14 @@ void CoverageServer::computeCoveragePath()
       F2CCells travel_ring_cells;
       travel_ring_cells.addGeometry(travel_ring);
       route_cells = decomp_gen_->decompose(travel_ring_cells, goal->decomp_mode);
-      swath_cells = goal->generate_headland ?
-        headland_gen_->generateHeadlands(route_cells, goal->headland_mode) : route_cells;
+      if (!goal->generate_headland) {
+        swath_cells = route_cells;
+      } else {
+        // Headland already removed before decomposing; only carve the inter-cell corridor here.
+        const double route_w = goal->route_headland_width > 0.0 ?
+          goal->route_headland_width : robot_params_->getOperationWidth();
+        swath_cells = headland_gen_->generateHeadlandsBetweenCells(route_cells, route_w);
+      }
     } else {
       // No decomposition: remove the headland from the single field cell
       if (goal->generate_headland) {
@@ -316,7 +321,7 @@ void CoverageServer::computeCoveragePath()
     visualizer_->visualize(
       field, field_no_headland, master_field.getRefPoint(),
       util::toCartesianNavPathMsg(path, header, path_gen_->getTurnPointDistance()),
-      swaths, header, headland_path);
+      swaths, header, headland_path, swath_cells);
     action_server_->succeeded_current(result);
   } catch (CoverageException & e) {
     RCLCPP_ERROR(get_logger(), "Invalid mode set: %s", e.what());
--- a/opennav_coverage/src/coverage_server.cpp
+++ b/opennav_coverage/src/coverage_server.cpp
@@ -48,6 +48,10 @@ CoverageServer::on_configure(const rclcpp_lifecycle::State & /*state*/)
     node, "default_generate_decomp", rclcpp::ParameterValue(false));
   get_parameter("default_generate_decomp", default_generate_decomp_);
 
+  nav2::declare_parameter_if_not_declared(
+    node, "default_generate_headland_swaths", rclcpp::ParameterValue(false));
+  get_parameter("default_generate_headland_swaths", default_generate_headland_swaths_);
+
   // If in GPS coordinates, we must convert to a CRS to compute coverage
   // Then, reconvert back to GPS for the user.
   nav2::declare_parameter_if_not_declared(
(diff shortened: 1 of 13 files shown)


opennav_coverage/src/coverage_server.cpp
```diff
@@ -213,8 +223,14 @@ void CoverageServer::computeCoveragePath()
...
       route_cells = decomp_gen_->decompose(travel_ring_cells, goal->decomp_mode);
-      swath_cells = goal->generate_headland ?
-        headland_gen_->generateHeadlands(route_cells, goal->headland_mode) : route_cells;
+      if (!goal->generate_headland) {
+        swath_cells = route_cells;
+      } else {
+        // Headland already removed before decomposing; only carve the inter-cell corridor here.
+        const double route_w = goal->route_headland_width > 0.0 ?
+          goal->route_headland_width : robot_params_->getOperationWidth();
+        swath_cells = headland_gen_->generateHeadlandsBetweenCells(route_cells, route_w);
```
SteveMacenski, line 232: Why not use or update `generateHeadlands`? You just created another API that probably ophans that one's call

SteveMacenski (commented):
This is a bit too large, I'm losing the thread on it and new APIs added that I'm not sure are required. The .action message is getting very crowded too.

opennav_coverage/src/coverage_server.cpp
u33549: The API I added generates headlands **between two cells**, not **within** a cell — `generateHeadlands` insets a headland ring around a cell's *entire* border, whereas `generateHeadlandsBetweenCells` only carves a corridor on the border **shared by two adjacent cells**.

There's currently no in-code caller of the `generateHeadlands(F2CCells, ...)` overload, but some tests are written against it, so I left the overload in place rather than removing it.

```
generateHeadlands(F2CCells)  —  headland inset around EACH cell's full border

   ┌───────────────┐   ┌───────────────┐
   │▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒│   │▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒│
   │▒┌───────────┐▒│   │▒┌───────────┐▒│
   │▒│           │▒│   │▒│           │▒│
   │▒│     A     │▒│   │▒│     B     │▒│
   │▒│           │▒│   │▒│           │▒│
   │▒└───────────┘▒│   │▒└───────────┘▒│
   │▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒│   │▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒│
   └───────────────┘   └───────────────┘
        ▒ = headland ring on ALL sides of every cell


generateHeadlandsBetweenCells  —  corridor ONLY on the shared A│B border

   ┌───────────────┬───┬───────────────┐
   │               │░░░│               │
   │               │░░░│               │
   │       A       │░░░│       B       │
   │               │░░░│               │
   │               │░░░│               │
   └───────────────┴───┴───────────────┘
                    ░ = carved corridor, only where two cells meet
                        (outer field boundary is left untouched)
```

u33549:
The `.action` crowding here comes from a single feature: **headland driving**. Normally the headland ring is only used as a turn/travel margin — the vehicle never actually covers it. This feature makes covering the headland ring itself optional, so the vehicle can drive it as part of the coverage plan instead of treating it as dead space.


It adds three goal fields:

- **`generate_headland_swaths`** — turns the feature on. When `false` (default), behavior is unchanged: the headland is only a margin, never driven.
- **`headland_first`** — orders the headland pass relative to the inner-field coverage: `true` drives the headland ring before the inner swaths, `false` drives it after.
- **`headland_full_coverage`** — chooses how much of the headland band gets covered: `false` is a single perimeter pass along the inner headland boundary; `true` sweeps the full band with F2C's `generateHeadlandSwaths`, using concentric passes spaced by the operation width (pass count = `headland_width / operation_width`) to fully cover the ring instead of just tracing its edge.

`headland_full_coverage=true` calls F2C v2's own `generateHeadlandSwaths` API, which already supports multiple concentric passes via its `n_swaths` parameter. `headland_full_coverage=false` instead uses `toHeadlandPerimeterPath`, a function we wrote ourselves — F2C has no built-in "single perimeter pass" mode, so that path and the toggle choosing between the two are our own additions.

image

The image below illustrates the concept this feature targets: the teal **Headland** band is normally just a turn/travel margin around the grey **Central field area** that the vehicle drives through but never covers. `generate_headland_swaths` makes that teal band coverable too, either with one lap along its inner edge or fully swept with concentric passes.

I considered defining a single pass as a special case of `n_swaths=1` instead of a separate toggle, but the two produce different geometries: our single pass traces the full headland width from the inner boundary (`field.buffer(-headland_width)`), while F2C's `n_swaths=1` only insets half an operation width from the outer boundary (`field.buffer(-operation_width * 0.5)`). Those only match if `headland_width == 0.5 * operation_width`, which isn't the normal case, so I kept it as an explicit toggle instead of collapsing it into the pass count.

u33549:
Also, if you'd like, I can split it up again regarding the length of pr

opennav_coverage/src/coverage_server.cpp
u33549: To add some context on why: the motivation was a real bug I ran into. Applying `generateHeadlands(F2CCells)` to every decomposed sub-cell insets a headland on **all** of that cell's borders, including borders shared with a neighbor. Two adjacent cells each carving their own headland on the same shared edge produced a double-width gap between them, and outer-boundary edges got a redundant *second* headland stacked on top of the one already removed at the field level (before decomposition).

`generateHeadlandsBetweenCells` only carves an edge when both its endpoints and midpoint lie on a neighbor cell's border, so a shared border gets exactly one corridor, and edges facing the outer boundary or a void keep their single, already-correct headland untouched.

SteveMacenski:
Yeah lets split this into individual features please. If any graphics help explain, please provide them!

PR #135, opened 2026-07-21
Carve route corridors only between adjacent decomposed cells

u33549:
## Summary

Fixes a double-headland gap between decomposed cells. Behavior-preserving unless
the field is decomposed with a headland.

**The bug:** when a field is decomposed with a headland, the headland was being
applied twice. The field headland is already removed once before decomposing;
then applying `generateHeadlands` to every sub-cell insets a headland on **all**
of that cell's borders again — including the borders shared with a neighbor. Two
adjacent cells each carving a headland on the same shared edge leaves a
double-width gap between them, and outer edges get a redundant second headland
stacked on the one already removed at the field level.

**The fix:** `generateHeadlandsBetweenCells` carves a corridor **only** on the
borders two cells actually share. An edge is carved only when both its endpoints
and its midpoint lie on a neighbor cell's border, so shared borders get exactly
one corridor and edges facing the outer boundary or a void keep their single,
already-correct headland. `route_headland_width` (default `0.0` = operation
width) sets the corridor width.

```
generateHeadlands per sub-cell (before)      generateHeadlandsBetweenCells (after)
headland inset on EVERY border               corridor ONLY on the shared A│B border

 ┌─────────────┐ ┌─────────────┐              ┌─────────────┬───┬─────────────┐
 │▒▒▒▒▒▒▒▒▒▒▒▒▒│ │▒▒▒▒▒▒▒▒▒▒▒▒▒│              │             │░░░│             │
 │▒┌─────────┐▒│ │▒┌─────────┐▒│              │             │░░░│             │
 │▒│    A    │▒│ │▒│    B    │▒│              │      A      │░░░│      B      │
 │▒│         │▒│ │▒│         │▒│              │             │░░░│             │
 │▒└─────────┘▒│ │▒└─────────┘▒│              │             │░░░│             │
 │▒▒▒▒▒▒▒▒▒▒▒▒▒│ │▒▒▒▒▒▒▒▒▒▒▒▒▒│              │             │░░░│             │
 └─────────────┘ └─────────────┘              └─────────────┴───┴─────────────┘
   double headland on the shared edge           one corridor where cells meet;
   + redundant outer headland                   outer boundary left untouched
```

### DIFF
--- a/opennav_coverage_msgs/action/ComputeCoveragePath.action
+++ b/opennav_coverage_msgs/action/ComputeCoveragePath.action
@@ -7,7 +7,7 @@ bool generate_headland True
 bool generate_route True
 bool generate_path True
 
-# Decomposition only: corridor width on cells' shared borders. 0 = operation width.
+# Decomposition only: corridor width on cells' shared borders. 0 uses the operation width. Other values override.
 float32 route_headland_width 0.0
 
 # The field specification to use.
--- a/opennav_coverage_msgs/action/ComputeCoveragePath.action
+++ b/opennav_coverage_msgs/action/ComputeCoveragePath.action
@@ -7,6 +7,9 @@ bool generate_headland True
 bool generate_route True
 bool generate_path True
 
+# Decomposition only: corridor width on cells' shared borders. 0 = operation width.
+float32 route_headland_width 0.0
+
 # The field specification to use.
 # If using polygons, bounding polygon must be first, followed by inner cutouts
 # Both must specify if the data is cartesian or GPS coordinates
--- a/opennav_coverage/include/opennav_coverage/headland_generator.hpp
+++ b/opennav_coverage/include/opennav_coverage/headland_generator.hpp
@@ -74,6 +74,16 @@ public:
     const F2CCells & cells, const opennav_coverage_msgs::msg::HeadlandMode & settings);
 
   /**
+   * @brief Carve a route corridor only on the borders cells actually share,
+   *        leaving edges facing the outer boundary or a void untouched.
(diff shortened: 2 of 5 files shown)


u33549:
image

The image compares the behavior of route generation before and after the fix:

* **Before (`generateHeadlands` per sub-cell):** 
  When `generateHeadlands` is applied individually to every sub-cell, a headland inset is carved around **all** borders of each cell. As a result, adjacent cells carving headlands along their shared boundary produce a **double-width gap** (double-headland) between them. Additionally, outer boundaries end up with a redundant, second headland stacked on top of the initial field-level headland.

* **After (`generateHeadlandsBetweenCells`):** 
  With `generateHeadlandsBetweenCells`, a corridor is carved **only** along the borders that two cells actually share. Outer boundaries and edges facing voids remain untouched (retaining their single, correct field-level headland), eliminating the redundant gaps and double-insets.

> **Note:** The generated route in the "After" image appears slightly distorted/crooked; this routing issue will be fixed in the next PR.

opennav_coverage_msgs/action/ComputeCoveragePath.action
```diff
@@ -7,6 +7,9 @@ bool generate_headland True
 bool generate_route True
 bool generate_path True
 
+# Decomposition only: corridor width on cells' shared borders. 0 = operation width.
```
SteveMacenski:
```suggestion
# Decomposition only: corridor width on cells' shared borders. 0 uses the operation width. Other values override.
```

SteveMacenski (approved):

PR #141, opened 2026-07-27
Add optional headland driving (perimeter loop or full band coverage)

u33549:
## Summary

Adds an optional "drive the headland" mode: normally the headland is just a
margin the swaths leave alone, but some operations need to cover that ring area
too. When enabled, a headland pass is generated and spliced onto the coverage
path. Off by default — no behavior change unless the new goal fields are set.

## New goal fields (`ComputeCoveragePath`)

| Field | Default | Meaning |
|-------|---------|---------|
| `generate_headland_swaths` | `False` | Also cover the headland ring area |
| `headland_first` | `False` | Drive the headland before (`True`) or after (`False`) the inner coverage |
| `headland_full_coverage` | `False` | `False` = one perimeter pass; `True` = concentric passes sweeping the whole band |

Needs `generate_route` and `generate_path` (the pass is spliced onto the planned
path); otherwise it is skipped with a warning.

## How it works

- **Single pass** (`headland_full_coverage = False`): one loop along the
  headland's inner boundary (`toHeadlandPerimeterPath` over `field_no_headland`).
- **Full band** (`headland_full_coverage = True`): `generateHeadlandSwaths`
  produces concentric rings spaced by the operation width
  (`count = round(headland_width / operation_width)`, min 1), each emitted as a
  perimeter pass from outer to inner.
- Each pass starts at the boundary point nearest the route to minimize the
  connecting jump, and is emitted as `HL_SWATH` path states. `discretizeSwathLike`
  now densifies `HL_SWATH` too (F2C's `discretizeSwath` only split `SWATH`), so
  the headland pass is a dense, followable path with per-pose velocity/direction.
- The pass is prepended or appended to the coverage path per `headland_first`.

## Visualization

The headland pass is published as a dedicated `coverage_server/headland_swaths`
marker (distinct color) so it isn't confused with the identically-shaped
planning-field boundary.

### DIFF
--- a/opennav_coverage/include/opennav_coverage/coverage_server.hpp
+++ b/opennav_coverage/include/opennav_coverage/coverage_server.hpp
@@ -132,6 +132,7 @@ protected:
   std::unique_ptr<Visualizer> visualizer_;
   bool cartesian_frame_;
   bool default_generate_decomp_{false};
+  bool default_generate_headland_swaths_{false};
 };
 
 }  // namespace opennav_coverage
--- a/opennav_coverage/include/opennav_coverage/headland_generator.hpp
+++ b/opennav_coverage/include/opennav_coverage/headland_generator.hpp
@@ -15,6 +15,7 @@
 #ifndef OPENNAV_COVERAGE__HEADLAND_GENERATOR_HPP_
 #define OPENNAV_COVERAGE__HEADLAND_GENERATOR_HPP_
 
+#include <cmath>
 #include <vector>
 #include <string>
 
@@ -82,6 +83,19 @@ public:
    */
   F2CCells generateHeadlandsBetweenCells(
     const F2CCells & cells, double route_width);
+
+  /**
+   * @brief Generate concentric rings that fully sweep the headland band.
+   *        Pass count is derived from the headland width and the operation
+   *        width (round(width / operation_width), min 1).
+   * @param field Full field (before headland removal)
+   * @param operation_width Spacing between rings
+   * @param settings Action request information (headland mode/width)
+   * @return Rings ordered outer to inner; each F2CCells is one pass
+   */
+  std::vector<F2CCells> generateHeadlandSwaths(
+    const Field & field, double operation_width,
+    const opennav_coverage_msgs::msg::HeadlandMode & settings);
 
   /**
    * @brief Sets the mode manually of the Headland for dynamic parameters
--- a/opennav_coverage/include/opennav_coverage/utils.hpp
+++ b/opennav_coverage/include/opennav_coverage/utils.hpp
@@ -15,6 +15,7 @@
 #ifndef OPENNAV_COVERAGE__UTILS_HPP_
 #define OPENNAV_COVERAGE__UTILS_HPP_
 
+#include <cmath>
 #include <vector>
 #include <string>
 #include <algorithm>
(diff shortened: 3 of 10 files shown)


u33549:
Visual comparison of the headland driving modes on an 18×18 m field with a 4 m headland:

Off (baseline) — `generate_headland_swaths: false`
The headland stays an uncovered margin; only the inner area is swathed.
image

Single perimeter pass — `headland_full_coverage: false`
One loop is driven along the headland's inner boundary (red headland_swaths marker), spliced onto the coverage path.
image

Full band coverage — `headland_full_coverage: true`
Concentric passes spaced by the operation width sweep the entire headland band (count = round(headland_width / operation_width)), from outer to inner.
image


The pass is emitted as dense HL_SWATH states, so it's a followable path with per-pose velocity/direction.

As I mentioned in a previous pr, paths can sometimes overlap in the navigation path; I’ve encountered this issue a few times, but I’m not sure if it’s our fault or F2C’s. In any case, v1 is still usable and offers a useful feature. Once everything is done, I’ll work on this issue separately.

SteveMacenski:
How many more PRs do you have lined up?

u33549:
I think this is the final feature integration. After that, one for overlap fix and maybe one for readme and tutorial. Also for other distoros. I wondered why you asked?

SteveMacenski (approved):
You have added alot in these PRs; I think some documentation explaining these new options and example uses of the feature to visualize the outcomes would be necessary (as well as a simple "here's how you do the basic from v1" workflow)

u33549:
Don't worry, I will. I just want to finish the features and optimizations. I worked really these improvements and I'm sure there may be bugs in it that I missed. I will also deal with future issues related to these parts in the future.

u33549:
I believe that the documentation process will help me identify some issues.

u33549:
I'll check to see if there's anything else on my checklist that needs to be done. If you want, I can start documenting instead of solving the overlap problem later.

u33549:
Remaining work I have planned:

- [ ] **Path overlap fix** — on some fields the generated `nav_path` doubles back over ground it
  has already covered. I haven't isolated yet whether it comes from our stitching layer or from
  F2C's path planner.
  - [ ] Build a minimal, deterministic reproduction
  - [ ] Identify the origin, then fix on our side or report 

- [ ] **Expose route connections in the coverage message** — `F2CRoute` carries connections
  (`getConnections()`) next to the swath groups, but they never reach the action result:
  `toCoveragePathMsg(F2CRoute, ...)` only flattens `getVectorSwaths()`, and `PathComponents` has
  no field for them. With `generate_path=true` they aren't lost, but they are indistinguishable —
  emitted as `HL_SWATH` states and folded into `msg.swaths`, so a consumer can't tell an
  inter-cell bridge from a real swath, which matters for anything computing covered area. The BT
  node's `coverage_path` port already advertises its output as "swaths and route connections",
  so this closes that gap too.
  - [ ] Add `nav_msgs/Path[] connections` to `PathComponents` (additive, backward compatible)
  - [ ] Populate it from `getConnections()` in the route overload
  - [ ] Publish connections as their own RViz marker, separate from swaths

- [ ] **Decomposition extras** *(on request — not scheduled, I don't think these are important
  unless someone asks)*
  - [ ] Render decomposed sub-cells as their own RViz marker
  - [ ] Expose F2C's `DecompObjective` (currently default-constructed) through `DecompMode`

- [ ] **Micro-swath fragments** *(may share a root cause with the overlap issue above — I'll look
  at both together)* — very short leftover swaths at corridor and notch ends produce a
  "turn → fragment → turn" artifact. A `min_swath_length` filter would remove them, but I rejected
  it because it can leave a ~0.35 m strip uncovered, and coverage guarantee comes first.

- [ ] **Documentation** — the new options with example uses and visualized outcomes, plus a short
  "here's the basic v1 workflow" walkthrough.

PR #137, opened 2026-07-25
Route decomposed fields with per-cell TSP and entry-angle-aware stitching

u33549:
## Summary

Routes decomposed fields by solving TSP per sub-cell and stitching results with
optimized corridor transitions. Enables coverage planning on multi-cell fields
with correct per-cell ordering. Rejects non-TSP route modes early.

## The Problem

Decomposed fields split into multiple sub-cells that share borders. A global
route planner (F2C's genRoute) can't optimize across cells — it processes the
entire multi-cell geometry at once and produces poor orderings or OOM errors on
large fields. A per-cell approach lets each sub-cell be solved independently at
TSP scale, then stitched together with optimized corridor transitions.

## The Solution

**Per-Cell TSP with Optimized Entry/Exit Selection:**
- Each decomposed sub-cell's swaths are solved independently via TSP (avoiding
  the all-in-one F2C genRoute that OOMs on large multi-cell fields)
- Each cell **can be entered from two ends** (start or reversed-end), each with
  its own entry heading angle
- After per-cell TSP, cells are connected via nearest-neighbor selection:
  for the current cell's end point, pick the next unvisited cell AND entry end
  that minimize: `distance + swath_width * (heading_mismatch_penalty)`
  - This avoids sharp-angle S-maneuvers between cells
  - Each unvisited cell is evaluated both forward and reversed
- **Bridge paths** between cells run along the **travel-cell pair border graph**
  (not straight lines that could cut voids), so routing stays within valid regions
- Result: globally-ordered swaths across all cells with entry-angle-aware,
  border-aware transitions

**Stitch Turn Optimization:**
- **Heading-aware entry:** corridor entry point chosen to minimize heading
  mismatch (perpendicular entry preferred over sharp angles)
- **Spike filtering:** bridge polylines drop attachment points that double back
  at junctions (numerical artifacts from geometry operations)

**Path Emissions:**
- Route connections (inter-cell jumps, boundary transitions) are now emitted as
  HL_SWATH states in nav_path so visualization and debugging tools can render
  the stitch points

**Early Validation:**
- Non-TSP route modes (BOUSTROPHEDON, SPIRAL, SNAKE, CUSTOM) rejected before
  geometry work; clear error message guides user to TSP or disable decomposition

## Decomposition Connection: PR #135

Route corridors between decomposed cells ([PR #135](https://github.com/open-navigation/opennav_coverage/pull/135))
fixed the double-headland gaps. This PR builds on that foundation by routing
[...]

PR #145, opened 2026-08-01
Support non-TSP route modes with field decomposition

u33549:
### Motivation

BOUSTROPHEDON, SNAKE, and SPIRAL were rejected whenever decomposition produced
multiple cells. The restriction came from #119, before this package had its own
per-cell stitching layer; #137 built that layer for TSP and the rejection was
never revisited. Since `default_route_type` is BOUSTROPHEDON, enabling
`generate_decomp` with stock parameters threw `INVALID_MODE_SET`.

### Changes

**Enable the swath-order modes for multi-cell fields.** Flattening the per-cell
swaths before ordering collided their per-cell ids and interleaved the cells, so
each cell is now ordered on its own and handed to the same stitch layer TSP uses
(extracted from `TspRouteMethod`). CUSTOM stays rejected — its order vector
indexes swaths absolutely and cannot be split across cells.

Two connection bugs surfaced while validating this in RViz. Both predate the
branch and both affect TSP as well:

**Cell-internal transitions ignored the boundary.** Each cell's swaths were
wrapped in a single connection-less group, so every transition became a naked
Dubins turn and a SNAKE/SPIRAL skip cut across covered ground. Cells now route
through F2C's shortest-path graph, connecting every consecutive swath pair.

**The detour test used a length ratio.** A connection collapsed to a single turn
below 1.3x the direct distance, which cannot separate a u-turn from a detour —
rounding a long cell's corner is only ~20% longer than cutting its diagonal. One
measured 36.3 m of polyline against 30.7 m direct and became a 30 m diagonal.
Peak deviation from the straight hop is now the criterion.

**Bridges trusted a half-resolved cell pair.** `getCellWherePoint` can miss an
endpoint sitting on a cell border. The graph was still built from the single
remaining cell and reached the far endpoint by jumping off its border. Since the
result had ≥2 points it counted as a success and the whole-set fallback never ran.

### Testing

Unit tests cover multi-cell ordering, SNAKE/SPIRAL planning, CUSTOM rejection,
and the unchanged single-cell path. Verified in RViz on concave E- and
zigzag-shaped fields across all modes.

### DIFF
--- a/opennav_coverage/include/opennav_coverage/route_generator.hpp
+++ b/opennav_coverage/include/opennav_coverage/route_generator.hpp
@@ -124,18 +124,18 @@ public:
   }
 
   /**
-   * @brief Whether the request (or the default it falls back to) is TSP.
-   *        Used to reject non-TSP + decomposition before any geometry work.
-   * @param settings Action request information
-   * @return true when the route mode resolves to TSP
+   * @brief Whether the goal's route mode resolves to CUSTOM, the one mode whose
+   *        absolute order vector cannot be split across decomposed cells
+   * @param settings Route mode settings from the action goal
+   * @return True if the effective mode is CUSTOM
    */
-  bool resolvesToTsp(const opennav_coverage_msgs::msg::RouteMode & settings)
+  bool resolvesToCustom(const opennav_coverage_msgs::msg::RouteMode & settings)
   {
     RouteType type = toType(settings.mode);
     if (type == RouteType::UNKNOWN) {
       type = default_type_;
     }
-    return type == RouteType::TSP;
+    return type == RouteType::CUSTOM;
   }
 
   /**
--- a/opennav_coverage/include/opennav_coverage/route_method.hpp
+++ b/opennav_coverage/include/opennav_coverage/route_method.hpp
@@ -59,16 +59,19 @@ public:
 /**
  * @class SwathOrderMethod
  * @brief Adapts the F2C swath-ordering modes (BOUSTROPHEDON, SNAKE, SPIRAL,
- *        CUSTOM). Flattens the per-cell swaths, orders them with the wrapped
- *        `SingleCellSwathsOrderBase`, and wraps the ordered swaths into a
- *        single-group `F2CRoute` (no connections) so the output type matches TSP.
+ *        CUSTOM). Single-cell input orders the flattened swaths with the wrapped
+ *        `SingleCellSwathsOrderBase`. Multi-cell input is ordered per cell and
+ *        stitched with the shared stitch layer (see route_method.cpp), the same
+ *        one `TspRouteMethod` uses. CUSTOM is rejected for multi-cell: its order
+ *        vector cannot be split across cells.
  */
 class SwathOrderMethod : public RouteMethod
 {
 public:
   SwathOrderMethod(
-    RouteType type, std::shared_ptr<f2c::rp::SingleCellSwathsOrderBase> orderer)
-  : type_(type), orderer_(std::move(orderer)) {}
+    const rclcpp::Logger & logger, RouteType type,
+    std::shared_ptr<f2c::rp::SingleCellSwathsOrderBase> orderer)
+  : logger_(logger), type_(type), orderer_(std::move(orderer)) {}
 
   F2CRoute plan(
     const F2CCells & travel_cells,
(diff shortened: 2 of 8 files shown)


u33549:
### Visual verification — concave E-shaped field

Decomposed into 4 cells (11 swaths each). All four route modes plan without
cutting across covered ground or leaving the field.

<details>
<summary>Goal used</summary>

```bash
ros2 action send_goal /compute_coverage_path opennav_coverage_msgs/action/ComputeCoveragePath "{
  generate_decomp: true, generate_headland: true, generate_route: true, generate_path: true,
  use_gml_file: false, frame_id: 'map',
  polygons: [{coordinates: [
    {axis1: 0.0, axis2: 0.0}, {axis1: 32.0, axis2: 0.0}, {axis1: 32.0, axis2: 8.0},
    {axis1: 8.0, axis2: 8.0}, {axis1: 8.0, axis2: 16.0}, {axis1: 32.0, axis2: 16.0},
    {axis1: 32.0, axis2: 24.0}, {axis1: 8.0, axis2: 24.0}, {axis1: 8.0, axis2: 32.0},
    {axis1: 32.0, axis2: 32.0}, {axis1: 32.0, axis2: 40.0}, {axis1: 0.0, axis2: 40.0},
    {axis1: 0.0, axis2: 0.0}]}],
  decomp_mode: {mode: 'TRAPEZOIDAL', split_angle: 1.5708},
  route_headland_width: 0.5,
  headland_mode: {mode: 'CONSTANT', width: 0.3},
  swath_mode: {objective: 'NUMBER', mode: 'BRUTE_FORCE'},
  route_mode: {mode: 'SPIRAL', spiral_n: 3},
  path_mode: {mode: 'DUBIN', continuity_mode: 'CONTINUOUS', turn_point_distance: 0.1}
}"
```

Only `route_mode` differs between the runs below.

</details>

**BOUSTROPHEDON**

BOUSTROPHEDON

**SNAKE** — note the skip pattern: consecutive passes sit two swath widths
apart, which is what the order modes offer over TSP and what TSP's
distance-only cost cannot produce.

SNAKE

**SPIRAL** (`spiral_n: 3`)

SPIRAL

**TSP** (`tsp_time_limit: 3`) — unchanged behaviour, included as the reference.

TSP

u33549:
### Known limitation: unsmoothed corners on border-following connections

Worth flagging since this PR makes it visible. When a connection follows the
border graph, the polyline is emitted verbatim as `HL_SWATH` states with no
curvature smoothing, so a right-angle corner in the cell boundary becomes a
right-angle corner in the path:

```cpp
PathState s;
s.point = pts[i];
s.angle = std::atan2(dy, dx);   // raw segment heading
s.type = f2c::types::PathSectionType::HL_SWATH;
```

corner 1
corner 2
corner 3

The rounded loops in these crops are `createTurn` output and respect the turning
radius; only the sharp vertices come from the polyline.

This is pre-existing behaviour from #137, not new here — but the old 1.3x length
ratio collapsed most of these connections into straight lines, so they rarely
surfaced. Switching to the deviation criterion keeps them, which is correct
geometrically and exposes the missing smoothing.

The corners are not followable at `min_turning_radius` (0.35 m in the demo), so
the controller will cut them. That is harmless where the cut falls inside the
cell, less so where the polyline hugs the field boundary.

I'm not sure this is worth fixing, or whether it causes a real problem in
practice — the controller may absorb it fine. Interested in your read. Unless
you say otherwise, I'll address it in a follow-up PR.

u33549:
Although this deviates from the plan in PR #141, it is one of the steps in resolving the overlap issue.

SteveMacenski (approved):

SteveMacenski:
What's the fix to have the smoothing applied in that edge case?

u33549:
I'll apply a minimum turn angle to these corners, or I'll create them as `turn` objects within F2C.

u33549:
I just want to get some feedback. What do you think about how we are going?

SteveMacenski:
Good in general! this is just alot of changes very fast in a piece of code I don't use often, so I'm kind of just letting it go right now with your best judgement except if I see something in particular I think needs to be changed. I'm trusting you to check your work and test edge cases that what you're doing works well robustly :-)

u33549:
Thank you I'm trying to do my best.

u33549:
image
image
image
image
image

I've done some work on anti-aliasing, and I'm a little confused, but these are the best results I've been able to get.

SteveMacenski:
Those seem good to me

PR #149, opened 2026-08-04
Follow and smooth the connections between swath groups

u33549:
### Problem

A connection between two swath groups was always driven as a single turn from the end of one group to the start of the next, ignoring the connection's own waypoints.

That is right for the u-turn between neighbouring swaths, where the track *is* the straight hop. It is wrong everywhere else. On a detour around a concave notch, or between decomposed cells, the route deliberately walks the border to stay inside the field, and planning a turn straight across it cuts through exactly the ground the route was avoiding.

### Approach

Follow the connection's own track, rounding its corners through the turn planner and leaving a corner square when no maneuver fits within half an operation width of it.

Which of the two treatments a connection gets is decided by **how far its track strays from the straight hop**, not by the ratio of their lengths. A long cell's rounded corner is only about 20% longer than the diagonal across it, yet shares no ground with it, so a length ratio cannot tell them apart.

The corners at either end are planned from the **headings of the swaths there**, not from the track's first and last legs. The border graph splices the swath endpoint into the ring, so those legs are often a short spur that the robot is not driving along. Reading the heading off them plans the turn from a pose the robot is never in: a 90° corner comes out as an S, a reversal as an L, and the arc meets the swath at a cusp. Corners too close together to be taken one at a time are folded into a single maneuver for the same reason.

### Known limitations

Two things I deliberately left out of this PR rather than hide:

1. **`kCornerCutTol` is a compile-time constant** (`0.5 × operation_width`). It is really a user-facing tolerance — how far a rounded corner may leave the track — and belongs as a ROS parameter beside `operation_width` and `min_turning_radius`. Kept as a constant here to hold the PR to one concern; happy to add the parameter here instead if you would rather it landed together.
2. **The end-heading logic and the shallow-corner bound are not unit tested.** The five tests above cover the rounding; the heading-aware ends, the corner folding and the jog bound were verified in RViz only.

The turn maths is heuristic and carries a number of tuning constants — the block at the top of `path_generator.cpp`. They are all fractions of the operation width or multiples of the turning radius, never absolute distances, but they are tuned rather than derived and are the part most worth a second opinion.

### DIFF
--- a/opennav_coverage/include/opennav_coverage/utils.hpp
+++ b/opennav_coverage/include/opennav_coverage/utils.hpp
@@ -16,6 +16,7 @@
 #define OPENNAV_COVERAGE__UTILS_HPP_
 
 #include <cmath>
+#include <limits>
 #include <vector>
 #include <string>
 #include <algorithm>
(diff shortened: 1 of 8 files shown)


u33549:
image
image
image
image
image
image
image


Here are some tests. I literally cooked something here.

u33549:
I know it's a really long pr, but since it forms a meaningful whole, I couldn't split it up.

opennav_coverage/include/opennav_coverage/utils.hpp
```diff
@@ -35,6 +36,147 @@ namespace opennav_coverage
...
+ *        Names a turn the same way in the debug log and in RViz
+ * @param idx Index of the turn
+ * @return Label
+ */
+inline std::string turnLabel(size_t idx)
+{
+  std::string label;
+  do {
+    label.insert(label.begin(), static_cast<char>('A' + idx % 26));
+    idx = idx / 26;
```
SteveMacenski, line 50: This is a bit unusual. Why not simply number them like everything else?

opennav_coverage/include/opennav_coverage/utils.hpp
```diff
@@ -35,6 +36,147 @@ namespace opennav_coverage
...
+/**
+ * @brief Magnitude of the turn from one heading to another, wrapped to [0, pi]
+ * @param from Heading turned from
+ * @param to Heading turned to
+ * @return Turn magnitude
+ */
+inline double sweepBetween(double from, double to)
+{
+  const double d = to - from;
+  return std::fabs(std::atan2(std::sin(d), std::cos(d)));
```
SteveMacenski, line 64: I think `angles` lib does this

SteveMacenski (commented):
I don't really understand this PR. 

Why is this happening in the first place and isn't that a bug that should be resolved within F2C APIs instead of creating our own thing here? This is no longer simply a wrapper on F2C for ROS/Nav2 but definiting its own capabilities which is not the intent of the project. It seems to me that there's either a bug in F2C or we're not using it correctly causing an issue (or is this issue not really an issue at all but a misuse of APIs?)

opennav_coverage/include/opennav_coverage/utils.hpp
u33549: In Rviz, the numbers sometimes appear side by side; to separate them, I used letters instead of numbers for debugging purposes, but I can remove this feature.

opennav_coverage/include/opennav_coverage/utils.hpp
u33549: i will check

u33549:
## There is a genuine F2C bug, and it's still open upstream

`f2c::pp::PathPlanning` already has exactly this capability — `planPathForConnection`, backed by `simplifyConnection`, with a signature that matches our `appendConnection` almost parameter-for-parameter: `(robot, s1, mp, s2, turn)`.

The reason we stopped calling it, back in [open-navigation/opennav_coverage#137](https://github.com/open-navigation/opennav_coverage/pull/137) ("Follow route connections in the path when they detour"), is this line in `simplifyConnection`:

```cpp
if (p1.distance(p2) < 6.0 * safe_dist || mp.size() < 2) {
  path.emplace_back(p2, ang2);   // the polyline is discarded here
  return path;
}
```

It decides whether to follow the connection's polyline or cut straight across it by the **straight-line distance between its endpoints alone** — never by how far the polyline in between actually strays from that line. A detour around a concave notch or between decomposed cells routinely has its endpoints within `6 * safe_dist` of each other while the path between them runs for tens of meters around real geometry. When that happens, `simplifyConnection` throws the polyline away and plans a straight hop through the void.

I checked this isn't fixed in the meantime: it's identical in the pinned v2.0.0 and in the current upstream tip (the in-progress v3 rework touches everything around it but leaves this exact condition untouched).

I'd also tried F2C directly before working around it, not just read the source — the results were poor on exactly the fields this matters for, concave and decomposed ones, which is what sent me looking for why.

## The heading bugs were ours, not F2C's — but going back isn't the fix either

The bugs this review caught — the S-shape, the teardrop, the wrong L — all trace back to one thing: our version read the heading at each end of a connection off the polyline's own first and last leg. `simplifyConnection` never does that; it takes `ang1`/`ang2` as parameters from the calling swaths and never infers them from the track. That part didn't need reinventing, and reinventing it is what let the bug in.

That doesn't make `simplifyConnection` the fix, though. It has real gaps of its own — at `vp.size() == 2` its fallback guard misses by one and drops both corners to a straight line, and its fillet loop skips the first and last corner it detects entirely, at any size. It also never checks what `createTurn` hands back for a loop or excess deviation.

So the corner logic here does more than `simplifyConnection` ever did, not less. The heading bug was ours, and is fixed; the rest was never F2C's to have covered.


## Why this isn't an isolated call

The corridor routing from an earlier PR diverges from F2C's own approach for the same kind of reason, and the per-cell TSP routing before that replaced F2C's global `genRoute` because the all-pairs route matrix it builds runs out of memory on a multi-cell field. Neither is a matter of using the API wrong — they're scaling and geometry limits hit on real fields, worked around one at a time. Both were built the same way: use F2C where it holds, and patch the specific place it doesn't, rather than stand up a parallel implementation next to it. This PR is the one where the patch grew past that.

## Where these patches should live

If it keeps this package from turning into a shadow of F2C, I'd rather fixes like this one not live inside `opennav_coverage` at all. A separate, general-purpose repo for F2C patches and extensions — this connection issue, the per-cell TSP work, whatever comes next — would keep this package thin and let each patch stand on its own rather than sit embedded in a robotics stack.

SteveMacenski:
What is the reason a PR can't be opened against F2C to resolve this issue and we target that newer verssion once merged?

u33549:
I open the upstream issue (and a PR if they want it), and this workaround stays here until that fix is released and we've migrated to the version that has it, then we drop this branch.

SteveMacenski:
Can you link the ticket/PR here so this is all cross referenced? 

I'd prefer to hold off until we hear back on that (or if we don't hear for a very long time)

u33549:
It's done. I'd appreciate it if you could take a look when you have a chance.

u33549:
Hi @SteveMacenski, 
The part needed for this PR is almost complete, but I’ll also move the changes we made here earlier—which actually belong in f2c—over there and open a cleanup PR for the codebase.

